PR #1, opened 2025-05-13
Consistent spacing on README

AvicennaJr:
A minor change but I have made the spacing and lines consistent for better display of the README file


superuser-does:
I personally found it easier to read as-is. The indentation of links naturally provides a visual hierarchy. This is especially when opened in a text editor instead of GitHub, which doesn't colour & underline links for example.

PR #17, opened 2026-01-06
Pinch prototype

fmasalha:


shtrom:
@fmasalha you'll need to edit your PR to point to the `autoland` branch (see bug 2007050).

PR #23, opened 2026-01-14
Bug 2003511 - Cancel loading bookmarks if weve navigated from the BookmarksFragment

boek:
I also moved the initial folder state to the BookmarksState instead of loading it directly in the middleware. Without the change it prevented us from navigating back on long loads.

### DIFF
--- a/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksMiddleware.kt
+++ b/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksMiddleware.kt
@@ -10,6 +10,7 @@ import androidx.navigation.NavController
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
 import mozilla.appservices.places.BookmarkRoot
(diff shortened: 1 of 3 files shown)


mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksMiddleware.kt
```diff
@@ -72,10 +73,11 @@ internal class BookmarksMiddleware(
     private val saveBookmarkSortOrder: suspend (BookmarksListSortOrder) -> Unit,
     private val lastSavedFolderCache: LastSavedFolderCache,
     private val reportResultGlobally: (BookmarksGlobalResultReport) -> Unit,
+    private val lifecycleScope: CoroutineScope,
     private val ioDispatcher: CoroutineDispatcher = Dispatchers.IO,
 ) : Middleware<BookmarksState, BookmarksAction> {
 
-    private val scope = CoroutineScope(ioDispatcher)
+    private val scope = CoroutineScope(lifecycleScope.coroutineContext + ioDispatcher)
```
MatthewTighe, line 80: neat

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksState.kt
```diff
@@ -113,7 +113,7 @@ internal data class BookmarksState(
             sortMenuShown = false,
             sortOrder = BookmarksListSortOrder.default,
             recursiveSelectedCount = null,
-            currentFolder = BookmarkItem.Folder("", "", null),
+            currentFolder = BookmarkItem.Folder("Mobile Bookmarks", BookmarkRoot.Mobile.id, null),
```
MatthewTighe: I feel a bit squeamish using the string literal `"Mobile Bookmarks"` here. We could resolve the string resource during construction of the Store in the Fragments instead [here](https://searchfox.org/firefox-main/source/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarkFragment.kt#92) and in EditBookmarkFragment.

Alternatively, we could do something like `val empty = <current default>` and introduce `fun Context.default() = BookmarksState(... currentFolder = <resolved string>)`. 

I think something is worth doing here but won't block on it

MatthewTighe (approved):
I would prefer to use a resolved string resource in place of the string literal you've added to the State default, but not strongly enough to block on it. LGTM otherwise

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksState.kt
boek: I ended up doing a check in the middleware that gives us the behavior we want. Ultimately this will get improved once we can tease apart the `BackClicked` action.

PR #33, opened 2026-01-21
Bug 2003434 - Fallback to mobile bookmarks root if last saved folder does not exist

segunfamisa:
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=2003434
Try: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=174138

### DIFF
--- a/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/menu/middleware/MenuDialogMiddleware.kt
+++ b/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/menu/middleware/MenuDialogMiddleware.kt
@@ -223,9 +223,16 @@ class MenuDialogMiddleware(
         val selectedTab = browserMenuState.selectedTab
         val url = selectedTab.getUrl() ?: return@launch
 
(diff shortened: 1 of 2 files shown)


boek (approved):

boek:
Are there any other places that we may need to do this?

segunfamisa:
> Are there any other places that we may need to do this?
I'm not really sure. I mostly based this PR on MarkH's work on phab and I added tests.

In my opinion, all of this dance should already be part of our "AddBookmarks" usecase. There's no reason for last saved cache to be used by the menu in order to determine where to save things. That should be part of our flow to add bookmarks.

And that would solve for "other places we should do this".

I have generally found it less than ideal that we have usecases, and we still reference and use the storage layers directly in various places.

I figured this required a slightly more involved refactoring that I wish we would do after we land this quick fix.

What are your thoughts? @boek  @MatthewTighe

boek:
That sounds great, let's add a follow up and put it [on our list](https://mozilla-hub.atlassian.net/wiki/spaces/Mobile/pages/2178252820/Technical+Investments).

PR #34, opened 2026-01-21
Bug 2003511 - Cancel loading bookmarks if weve navigated from the Bookmarks Fragment

boek:
[try](https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=174243)

### DIFF
--- a/mobile/android/fenix/app/src/test/java/org/mozilla/fenix/bookmarks/BookmarksMiddlewareTest.kt
+++ b/mobile/android/fenix/app/src/test/java/org/mozilla/fenix/bookmarks/BookmarksMiddlewareTest.kt
@@ -9,7 +9,9 @@ import androidx.navigation.NavBackStackEntry
 import androidx.navigation.NavController
 import androidx.navigation.NavDestination
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.StandardTestDispatcher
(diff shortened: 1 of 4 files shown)


mobile/android/fenix/app/src/test/java/org/mozilla/fenix/bookmarks/BookmarksMiddlewareTest.kt
```diff
@@ -2064,6 +2068,7 @@ class BookmarksMiddlewareTest {
         saveBookmarkSortOrder = saveSortOrder,
         lastSavedFolderCache = lastSavedFolderCache,
         reportResultGlobally = reportResultGlobally,
+        lifecycleScope = CoroutineScope(testDispatcher),
```
MatthewTighe, line 2071: These are different dispatchers, not sure if we care about that. I think this would mean the ioDispatcher overwrites the test dispatcher you've used here when we combine the contexts in the implementation

MatthewTighe (approved):
LGTM

mobile/android/fenix/app/src/test/java/org/mozilla/fenix/bookmarks/BookmarksMiddlewareTest.kt
MatthewTighe: I apparently have forgotten how to highlight a section in Github - this refers to the dispatcher argument provided for `ioDispatcher`

segunfamisa: is there a reason we did not use `TestScope()`  here?

PR #35, opened 2026-01-21
S2s initial branch files

fmasalha:


mobile/android/android-components/.buildconfig.yml
```diff
@@ -2480,3 +2480,32 @@ projects:
...
+      - components:tooling-lint
+  components:feature-summarization-wiring:
+    description: Wiring factory for assembling the Summarization (S2S) feature.
+    path: components/feature/summarization/wiring
+    publish: false
+    upstream_dependencies:
+      - components:feature-summarization
+      - components:feature-summarization-impl
+      - components:feature-summarization-api
+      - components:tooling-lint
```
jonalmeida: @fmasalha hey, quick drive-by 🚗 : for these API modules, we have categorizations similar to the ones you have here that fit together with our patterns instead of everything being a "feature" module. Have a look at our docs and let me know if you have question: https://mozac.org/contributing/architecture

For example, `feature-summarization-api` would probably be `concept-summarization`

fmasalha: Thank you for the info, this patch was updated to only include the feature flag with the structure being covered in https://github.com/mozilla-firefox/firefox/pull/37/commits/257d4279d6651d6b7d7f6fc1c59edfc32f653c8f

segunfamisa (approved):
Thanks @fmasalha this looks good to me.

Please don't forget to update the pull request title

PR #41, opened 2026-01-23
Bug 2010695 - Add empty modules for S2S feature

MatthewTighe:


mobile/android/android-components/components/concept/llm/README.md
```diff
@@ -0,0 +1,9 @@
+# [Android Components](../../../README.md) > Concept > LLM
+
+
```
segunfamisa: I think we have an opportunity to properly document and explain what this module is about. I think we should do that in the readme.

mobile/android/android-components/components/feature/summarize/README.md
```diff
@@ -0,0 +1,9 @@
+# [Android Components](../../../README.md) > Feature > SUMMARIZE
+
+
```
segunfamisa: Same here and other modules

segunfamisa (approved):
Thank you!

That looks good to me. My only comment is to use this opportunity to start with good enough READMEs so that we don't have to do it retroactively.

I think the same blobs we have in `buildconfig.yml` is even sufficient. What do you think? I'm approving anyway.

Thank you!

mobile/android/android-components/.buildconfig.yml
```diff
@@ -1702,6 +1719,15 @@ projects:
     publish: true
     upstream_dependencies:
     - components:tooling-lint
+  components:lib-llm-gemini-nano:
```
jonalmeida: nit naming: the hyphens are like namespaces. So you shouldn't need the last one. You've got the right format otherwise.

For example, `lib-fetch` has `httpurlconnection` and `okhttp` versions.

```suggestion
  components:lib-llm-gemininano:
```

mobile/android/android-components/components/concept/llm/README.md
MatthewTighe: Yes, this was just intended to be a placeholder for now. I've copied over the descriptions from the buildconfig, but let's flesh this out as we work on the modules themselves.

mobile/android/android-components/.buildconfig.yml
MatthewTighe: okay, thanks! is this documented anywhere?

boek (approved):

mobile/android/android-components/.buildconfig.yml
jonalmeida: It's a general package naming strategy. The google style guide has a decent description for it: https://google.github.io/styleguide/javaguide.html#s5.2.1-package-names

PR #46, opened 2026-01-27
Bug 2010700 - Created public compose root and store primitives for s2s

fmasalha:


mobile/android/android-components/components/feature/summarize/build.gradle
```diff
@@ -16,13 +17,24 @@ android {
...
 
     namespace = 'mozilla.components.feature.summarize'
 }
 
 dependencies {
+    implementation project(':components:lib-state')
+
     testImplementation libs.androidx.test.junit
+
+    implementation platform(libs.androidx.compose.bom)
```
MatthewTighe: Do we need this? I would think since we specify versions through `libs.versions.toml` that we do not

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationStore.kt
```diff
@@ -0,0 +1,18 @@
...
+
+class SummarizationStore(
+    initialState: SummarizationState,
+    reducer: Reducer<SummarizationState, SummarizationAction> = ::summarizationReducer,
+    middleware: List<SummarizationMiddleware> = listOf(),
+) : Store<SummarizationState, SummarizationAction>(
+    initialState,
+    reducer,
+    middleware,
+)
```
MatthewTighe: One thing @boek and I have chatted about is using `typealias`es for store definitions instead of defining new classes. The reasoning for this is it automatically defends against some anti-patterns we've seen and used - for example:
1. a `typealias` can't expect different constructor parameters
2. we can't define an `init` block and launch side-effects  

These are both on display in [BookmarksStore](https://searchfox.org/firefox-main/source/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksStore.kt#19).

This is optional feedback, but I think we should try it here. That would just look like:
`typealias SummarizationStore = Store<SummarizationState, SummarizationAction` and no changes should be required to the instantiations you've already got.

Since that becomes so short, I think we could probably also combine that with the `Action` and `State` definitions in one file.

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationScreen.kt
```diff
@@ -0,0 +1,56 @@
...
+            initialState = SummarizationState(
+                pageSummarizationState = PageSummarizationState.Inert(url),
+            ),
+            middleware = listOf(SummarizationMiddleware()),
+        ),
+    )
+}
+
+@Composable
+internal fun SummarizationScreen(store: SummarizationStore) {
```
MatthewTighe: I think this can be `private` as well

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationState.kt
```diff
@@ -0,0 +1,30 @@
...
+
+sealed class PageSummarizationState {
+    data class Inert(val url: String) : PageSummarizationState()
+    data class WaitingForPermission(val url: String) : PageSummarizationState()
+    data object Summarizing : PageSummarizationState()
+    data class Summarized(val text: String) : PageSummarizationState()
+    data class Error(val error: SummarizationError) : PageSummarizationState()
+}
+
+sealed class SummarizationError {
```
MatthewTighe: These should all be `internal` - we don't want them part of the public API

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationStore.kt
```diff
@@ -0,0 +1,18 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.summarize
+
+import mozilla.components.lib.state.Reducer
+import mozilla.components.lib.state.Store
+
+class SummarizationStore(
```
MatthewTighe: This should also be `internal` regardless of whether we change it to a typealias.

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationState.kt
```diff
@@ -0,0 +1,30 @@
...
+import mozilla.components.lib.state.State
+
+data class SummarizationState(
+    val pageSummarizationState: PageSummarizationState,
+    val summarizedText: String = "",
+) : State
+
+sealed class PageSummarizationState {
+    data class Inert(val url: String) : PageSummarizationState()
+    data class WaitingForPermission(val url: String) : PageSummarizationState()
```
MatthewTighe: optional: I think we could just leave these as `data objects` for now and remove the `url` parameter. It doesn't really matter, but it might "feel" more changeable to someone else down the road the more barren the definition is at first.

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationScreen.kt
```diff
@@ -0,0 +1,56 @@
...
+package mozilla.components.feature.summarize
+
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.tooling.preview.PreviewParameter
+import androidx.compose.ui.tooling.preview.PreviewParameterProvider
+
+@Composable
+fun SummarizationUi(url: String) {
```
MatthewTighe: optional: I think we could remove the `url` parameter. It doesn't really matter, but it might "feel" more changeable to someone else down the road the more barren the definition is at first.

MatthewTighe (approved):
Nice, this is looking good. **Make sure you run a try** - I imagine this will hit failures on linting for some of the undocumented public APIs at least. I don't think we have any real automation setup for Github yet. Other than that, I don't have any blocking feedback. Since we are planning to iterate on we can just make changes as we discover we need them

fmasalha:
Ty for the review, here is the try attempt https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=175715 

Ran the linter locally and got no errors after adding a docstring for the public compose function.

mobile/android/android-components/components/feature/summarize/build.gradle
fmasalha: Ye I saw it in another feature component and just assumed it was required. Thanks for pointing that out!

fmasalha:
updated try https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=175755

fmasalha:
hopefully the final updated try: https://treeherder.mozilla.org/jobs?repo=try&revision=4f6862874bbdb8ea13a3de49352b522db8b7a101

fmasalha:
https://treeherder.mozilla.org/jobs?repo=try&revision=892af32670ccc25595549c9b314cf3db394db615

All green try ^

PR #49, opened 2026-01-29
Bug 2010708 - Created SummarizationFragment in Fenix

fmasalha:


fmasalha:
There are some open issues that need in follow up's.

~~1. The menu button has a icon that i just chose randomly to fit the design, of course ui will probably give use guidance here and since this feature is only available in debug, i think this is fine for now but will need a follow up.~~ (Read comment below)

2. The button remains enabled/visable no matter what as there currently does not exist a way to tell if it should show/or be enabled other than the feature flag, this will need a followup since I imagine we need a button for debug purposes.

fmasalha:
>     1. The menu button has a icon that i just chose randomly to fit the design, of course ui will probably give use guidance here and since this feature is only available in debug, i think this is fine for now but will need a follow up.

So I saw them post the figma in the slack and i added the lightning symbol they used for the menu button removing this concern.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/menu/compose/MoreSettingsSubmenu.kt
```diff
@@ -38,7 +38,9 @@ internal fun MoreSettingsSubmenu(
     translationInfo: TranslationInfo,
     showShortcuts: Boolean,
     isAndroidAutomotiveAvailable: Boolean,
+    shakeToSummarizeFeatureEnabled: Boolean,
```
segunfamisa: I think that naming this something like "showSummarization" is a bit more future-proof, because very soon, we may not be only checking if the feature is enabled. We might need to know if the user has turned it off.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationFragment.kt
```diff
@@ -0,0 +1,40 @@
...
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View = ComposeView(requireContext()).apply {
+        setViewCompositionStrategy(ViewCompositionStrategy.DisposeOnViewTreeLifecycleDestroyed)
+        setContent {
+            FirefoxTheme {
+                SummarizationUi()
+            }
+        }
+    }
```
segunfamisa: Let's use the `content` API instead. It's essentially a shortcut for the `ComposeView()....setViewCompositionStrategy` part

```suggestion
    ): View = content {
        FirefoxTheme {
            SummarizationUi()
        }
    }
```

segunfamisa (approved):
@fmasalha this looks good to me. Thanks for the PR.

Can we please have a `./mach try --preset firefox-android` try before landing?

I have added a couple of possible improvements that we should consider. Let me know what you think

fmasalha:
> @fmasalha this looks good to me. Thanks for the PR.
> 
> Can we please have a `./mach try --preset firefox-android` try before landing?
> 
> I have added a couple of possible improvements that we should consider. Let me know what you think

Thanks for the review, seems that try attempts are broken rn https://mozilla.slack.com/archives/CRLK0EXEF/p1770130031403859 

Will submit one once its back up hopefully.

I was getting a LongMethod lint error on the MoreSettingsSubmenu after adding the menu button, I split it up in case you are wondering why those changes are there.

fmasalha:
try attempt: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=177233

fmasalha:
Try attempt: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=177279

PR #50, opened 2026-01-30
Bug 2011536 - sanitize intent before interacting with it.

pollym:
We are seeing `ClassCastException`s in the crash report - this is likely to be happening because some external intent has serialized some class that our app doesn't know about. Calling this `sanitize()` function strips out unknown classes from the bundle. `Intent.putExtra()` will internally call unparcel(), which assumes the bundle is serializable, so we need to call sanitize before doing this.

[Running a try here](https://treeherder.mozilla.org/jobs?repo=try&revision=9a328763a8bcbf33ecc7e9116aceaa5854f85157)


jonalmeida (approved):
This is pretty straight-forward fix to me, thanks!

nit: if the change is a speculative fix, can we add that to the commit message so that we know we can re-open the bug if it doesn't fix the all the crashes linked to it.

jonalmeida:
testing-exception-other: we ought to test this by monitoring our Nightly audience. Automated testing would not be able to uncover much.

PR #51, opened 2026-02-03
Bug 2011498 - Add an accelerometer API, implementation, and shake detection

MatthewTighe:


mobile/android/fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -107,6 +112,15 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
...
 
+        val accelerometerFlow = LifecycleAwareSensorManagerEventFlow((requireActivity().getSystemService(Context.SENSOR_SERVICE) as SensorManager))
+
+        lifecycle.addObserver(accelerometerFlow)
+        lifecycleScope.launch {
+            SampledAccelerometer(accelerometerFlow).detectShakes().collect {
+                Log.i("shaker", "shake detected")
+            }
+        }
+
```
MatthewTighe: I will remove this before landing, just wanted an example of the API

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
```diff
@@ -0,0 +1,69 @@
...
+
+private const val NUM_SAMPLE_REPLAY = 5
+
+/**
+ * This class is an adapter between Android platform accelerometer data and
+ * platform-agnostic data. It uses the Android [SensorManager] to collect this data,
+ * and can be added as a lifecycle observer to handle registering and unregistering its
+ * sensor management automatically.
+ */
+class LifecycleAwareSensorManagerEventFlow(
```
MatthewTighe: java.java.java

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
```diff
@@ -0,0 +1,69 @@
...
+            event.values[2],
+            event.timestamp,
+        )
+        _samples.tryEmit(sample)
+    }
+
+    override fun onResume(owner: LifecycleOwner) {
+        super.onResume(owner)
+        logger("Registering self as sensor listener")
+        sensorManager.registerListener(this, sensor, SensorManager.SENSOR_DELAY_UI)
```
MatthewTighe, line 62: We will follow-up on the sampling delay here - this may be too high of a rate for battery reasons and we can likely do something intelligent in terms of increasing the sample rate as necessary.

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
```diff
@@ -0,0 +1,69 @@
...
+) : AccelerometerEventFlow, SensorEventListener, DefaultLifecycleObserver {
+
+    private val sensor: Sensor? by lazy {
+        sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER)
+    }
+
+    private val _samples = MutableSharedFlow<Accelerometer.Sample>(
+        replay = NUM_SAMPLE_REPLAY,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST,
+    )
```
MatthewTighe, line 43: The current sample rate is too high for `BufferOverflow.SUSPEND`, which is the default. The backpressure seems to overwhelm the consumer, but we shouldn't really need to care about processing every value anyway. 

We can also reconsider this when we follow-up on the sample rate as commented below.

MatthewTighe (commented):
Couple notes in here for follow-ups. My main priority is to get this landing to unblock an end-to-end test of an early implementation.

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
```diff
@@ -0,0 +1,69 @@
...
+    override fun flowEvents(): Flow<Accelerometer.Sample> = _samples
+
+    override fun onAccuracyChanged(sensor: Sensor, accuracy: Int) = Unit
+
+    override fun onSensorChanged(event: SensorEvent) {
+        val sample = Accelerometer.Sample(
+            event.values[0],
+            event.values[1],
+            event.values[2],
+            event.timestamp,
```
segunfamisa: Let's consider using named params here. It's not immediately obvious which is what (esp when reading the code outside the IDE)



```suggestion
            xAccel = event.values[0],
            yAccel = event.values[1],
            zAccel = event.values[2],
            timestampNs = event.timestamp,
```

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
segunfamisa: ⚠️ nitpick :
I'm personally not the biggest fan of having the flow suffix because i'm not sure it adds that much information.
I also think we can drop the SensorManager because from a reader's POV, I already know an accelerometer is some kind of sensor...so it's also not adding that much information, and at the same time, it's making our name Java-ish 😂
I think we can call this `AndroidAccelerometerEventsSource` or `LifecyleAwareAccelerometerEventsSource`

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
```diff
@@ -0,0 +1,69 @@
...
+    private val sensor: Sensor? by lazy {
+        sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER)
+    }
+
+    private val _samples = MutableSharedFlow<Accelerometer.Sample>(
+        replay = NUM_SAMPLE_REPLAY,
+        onBufferOverflow = BufferOverflow.DROP_OLDEST,
+    )
+
+    override fun flowEvents(): Flow<Accelerometer.Sample> = _samples
```
segunfamisa: ❓ `question`:

Do we need to do anything if there's no accelerometer sensor? I think we may need to throw an error that will be handled downstream?

I don't know how we envisage the usage of the sample stream, but we could accidentally have an observer waiting indefinitely if we don't emit anything 🤔 

It sounds like we should complete it - either successfully empty or with an error, if we don't have a sensor

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
segunfamisa: Yeah. Makes sense. I think DROP_OLDEST is a totally sensible strategy 👌

segunfamisa, line 43: ❓ `question`:

I'm curious to learn why we need a replay here. Do we need new subscribers to see old values?

---
From my POV, the idea of such a high stream of data is like a firehose. If you missed the last x values, better wait for the next ones.

I think what we might need is to make the `extraBufferCapacity` non-zero (maybe like 32 or something), so that we can apply the `BufferOverflow.DROP_OLDEST` strategy.

From [the docs](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-shared-flow.html):
> Values other than [BufferOverflow.SUSPEND](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.channels/-buffer-overflow/-s-u-s-p-e-n-d/index.html) are supported only when replay > 0 or extraBufferCapacity > 0. 
> 
> Buffer overflow can happen only when there is at least one subscriber that is not ready to accept the new value. In the absence of subscribers only the most recent [replay](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/-mutable-shared-flow.html) values are stored and the buffer overflow behavior is never triggered and has no effect.

mobile/android/android-components/components/concept/accelerometer/src/main/java/mozilla/components/concept/accelerometer/Accelerometer.kt
```diff
@@ -0,0 +1,40 @@
...
+ * A flow of data of accelerometer events. This is separated from [Accelerometer] so that we
+ * can wrap platform-specific implementations of data streams separate from any behavior we want
+ * to be surfaced from the [Accelerometer] itself.
+ */
+fun interface AccelerometerEventFlow {
+    /**
+     * The flow of data.
+     */
+    fun flowEvents(): Flow<Accelerometer.Sample>
+}
```
segunfamisa: I feel like there is a bit of diversion in the naming in this type.
The interface is an event flow, the function is flow of events but the return type is flow of sample
I also think we can use consistent "sample" name, if that's the name we want to use.
Or we can rename Sample to Event, if we prefer to preserve the event nomenclature.

I think we could also consider naming the interface something like `AccelerometerEventSource` or `AccelerometerDataSource` (further comments on this in the implementation)

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * This class is an adapter between Android platform accelerometer data and
+ * platform-agnostic data. It uses the Android [SensorManager] to collect this data,
+ * and can be added as a lifecycle observer to handle registering and unregistering its
+ * sensor management automatically.
+ */
+class LifecycleAwareSensorManagerEventFlow(
+    private val sensorManager: SensorManager,
+    private val logger: (String) -> Unit = { message ->
+        Logger("mozac/LifecycleAwareSensorManagerEventFlow").info(message)
```
segunfamisa: 👌 neat

do we need info level logs here or debug is fine?

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/SampledAccelerometer.kt
```diff
@@ -0,0 +1,29 @@
...
+package mozilla.components.lib.accelerometer.sensormanager
+
+import android.hardware.SensorManager
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+import mozilla.components.concept.accelerometer.Accelerometer
+import mozilla.components.concept.accelerometer.AccelerometerEventFlow
+
+/**
+ *
```
segunfamisa: Missing docs?

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/test/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
segunfamisa, line 1: I think the file name should carry a `Test` suffix or something?

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/test/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
```diff
@@ -0,0 +1,72 @@
...
+
+        org.mockito.Mockito.verify(mockSensorManager).unregisterListener(eventFlow)
+        assertEquals(
+            listOf("Registering self as sensor listener", "Unregistering self as sensor listener"),
+            logMessages,
+        )
+    }
+}
+
+class FakeLifecycleOwner : LifecycleOwner {
```
segunfamisa: Just an FYI: We also have TestLifecycleOwner in https://developer.android.com/jetpack/androidx/releases/lifecycle#kotlin 

And that can help us to just trigger the lifecycle events without having to call the `onResume(...)` ourselves - since we won't be calling those methods manually in real life.

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/test/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
```diff
@@ -0,0 +1,72 @@
...
+    private lateinit var mockSensor: Sensor
+    private lateinit var eventFlow: LifecycleAwareSensorManagerEventFlow
+    private val logMessages = mutableListOf<String>()
+
+    @Before
+    fun setUp() {
+        mockSensorManager = mock()
+        mockSensor = mock()
+        whenever(mockSensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER)).thenReturn(mockSensor)
+
```
segunfamisa: I want to propose using robolectric shadows here instead of mocking to be able to write the tests properly and without mocks.

Here's a sample test I was able to get working:

```kotlin
@RunWith(AndroidJUnit4::class)
@Config(sdk = [28])
class LifecycleAwareSensorManagerEventFlowTest {

    private val testDispatcher = StandardTestDispatcher()

    // from AndroidX
    private lateinit var lifecycleOwner: TestLifecycleOwner

    private lateinit var sensorManager: SensorManager
    private lateinit var shadowSensorManager: ShadowSensorManager

    private lateinit var eventFlow: LifecycleAwareSensorManagerEventFlow

    @Before
    fun setUp() {
        val context = ApplicationProvider.getApplicationContext<Context>()
        sensorManager = context.getSystemService(Context.SENSOR_SERVICE) as SensorManager
        shadowSensorManager = Shadows.shadowOf(sensorManager)
        shadowSensorManager.addSensor(ShadowSensor.newInstance(Sensor.TYPE_ACCELEROMETER))

        eventFlow = LifecycleAwareSensorManagerEventFlow(
            sensorManager = sensorManager,
            logger = { },
        )

        lifecycleOwner = TestLifecycleOwner(Lifecycle.State.INITIALIZED)
        lifecycleOwner.lifecycle.addObserver(eventFlow)
    }

    @Test
    fun `when lifecycle is resumed, sensor events are emitted`() = runTest(testDispatcher) {
        // given the lifecycle is resumed
        lifecycleOwner.handleLifecycleEvent(Lifecycle.Event.ON_RESUME)

        // and we are observing the event flow
        val samples = mutableListOf<Accelerometer.Sample>()
        val eventsJob = launch(testDispatcher) {
            eventFlow.flowEvents()
                .collect { samples.add(it) }
        }

        // when we emit two sensor events
        emitSensorEvent(floatArrayOf(1f, 2f, 3f, 4f))
        emitSensorEvent(floatArrayOf(1f, 2f, 3f, 4f))

        testDispatcher.scheduler.advanceUntilIdle()

        // then we observe two emissions
        assertEquals("Two sensor events are observed", 2, samples.size)
        eventsJob.cancel()
    }

    @Test
    fun `when lifecycle is paused, no sensor events are emitted`() = runTest(testDispatcher) {
        // given that we are observing the events
        val samples = mutableListOf<Accelerometer.Sample>()
        val eventsJob = launch(testDispatcher) {
            eventFlow.flowEvents()
                .collect { samples.add(it) }
        }

        // and two events were emitted when the life cycle was resumed
        lifecycleOwner.handleLifecycleEvent(Lifecycle.Event.ON_RESUME)
        emitSensorEvent(floatArrayOf(1f, 2f, 3f, 4f))
        emitSensorEvent(floatArrayOf(1f, 2f, 3f, 4f))

        testDispatcher.scheduler.advanceUntilIdle()

        // when the lifecycle is paused
        lifecycleOwner.handleLifecycleEvent(Lifecycle.Event.ON_PAUSE)

        // and we emit another event
        emitSensorEvent(floatArrayOf(1f, 2f, 3f, 4f))

        testDispatcher.scheduler.advanceUntilIdle()

        // then we observe two emissions
        assertEquals("Expected that only the two prior sensor events were emitted", 2, samples.size)
        eventsJob.cancel()
    }

    private fun emitSensorEvent(values: FloatArray = floatArrayOf(1f, 2f, 3f, 4f)) {
        val sensor = sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER)
        shadowSensorManager.sendSensorEventToListeners(
            SensorEventBuilder.newBuilder()
                .setSensor(requireNotNull(sensor))
                .setValues(values)
                .setTimestamp(1234)
                .build(),
        )
    }
```

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/test/java/mozilla/components/lib/accelerometer/sensormanager/SampledAccelerometerTest.kt
```diff
@@ -0,0 +1,106 @@
...
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.take
+import kotlinx.coroutines.flow.toList
+import kotlinx.coroutines.test.runTest
+import mozilla.components.concept.accelerometer.Accelerometer
+import mozilla.components.concept.accelerometer.AccelerometerEventFlow
+import org.junit.Assert.assertEquals
+import org.junit.Test
+
+class SampledAccelerometerTest {
```
segunfamisa: 👌

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/SampledAccelerometer.kt
```diff
@@ -0,0 +1,29 @@
...
+ */
+class SampledAccelerometer(
+    private val flow: AccelerometerEventFlow,
+) : Accelerometer {
+
+    override fun gravityNormalizedSamples(): Flow<Accelerometer.Sample> = flow
```
segunfamisa: Are we going to have the sampling happen later? Or it has already happened somewhere? 😅

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
```diff
@@ -0,0 +1,91 @@
...
+        .map { Unit }
+
+/**
+ * Configuration element for the sensitivity of shake detection.
+ */
+enum class ShakeSensitivity {
+    Low,
+    Medium,
+    High,
+}
```
segunfamisa: What if we make this a class, and we provided Low, High and Medium values as this:

```kotlin

@JvmInline
value class ShakeSensitivity(val threshold: Float) {
  companion object {
    val Low = ShakeSensitivity(threshold = 3.2f)
    val Medium = ShakeSensitivity(threshold = 2.7f)
    ....
  }
}

```
The benefit of this idea is that it gives room for consumers of this class to be able to use their own in-between sensitivity thresholds without us having to change our API, and we don't have to manually map the given sensitivity into a threshold in the `next()` function besides the default ones we've defined.

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
```diff
@@ -0,0 +1,91 @@
...
+
+private data class ShakeState(
+    val hits: Int = 0,
+    val windowStartNs: Long = 0L,
+    val lastShakeNs: Long = 0L,
+    val shouldEmit: Boolean = false,
+) {
+    fun next(
```
segunfamisa, line 60: From a reader's POV, this function is a tough one to parse.

I think we could add some comments, explaining how the algorithm works, what we are trying to do

segunfamisa (changes requested):
Thanks for the PR.

This PR is definitely shaping up nicely, thank you.

I have some comments, I highlighted the ones that about naming that are nitpicky - so, please feel free to not pay too much attention to them.

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
MatthewTighe: How are you imagining that to look? We don't have a way to complete the flow on the publishing side, so I think this would mean that we'd need to change the `AccelerometerEventSource` to be a `Flow<SensorResult>` so that consumers could cancel their collection if we had a `SensorResult.Failed`. 

Is there anything they could do to recover from this case? For our specific use-case, I don't think the resulting behavior is different - we just don't get shake detection.

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
MatthewTighe: This makes more sense to me, thanks!

mobile/android/android-components/components/concept/accelerometer/src/main/java/mozilla/components/concept/accelerometer/Accelerometer.kt
```diff
@@ -0,0 +1,40 @@
...
+package mozilla.components.concept.accelerometer
+
+import kotlinx.coroutines.flow.Flow
+
+/**
+ * A flow of data of accelerometer events. This is separated from [Accelerometer] so that we
+ * can wrap platform-specific implementations of data streams separate from any behavior we want
+ * to be surfaced from the [Accelerometer] itself.
+ */
+fun interface AccelerometerEventFlow {
```
boek: As discussed in Slack this is functionally the same interface as `Accelerometer`. We can achieve the same level of testability while simplifying the number of types that we expose from these modules.

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
segunfamisa: Yeah. Let's keep this and see if we need to revisit

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/test/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerEventFlow.kt
MatthewTighe: I am not exactly sure why, but applying these tests for me was racy because the `launch`ed collection didn't start in time. I had to `runCurrent` on the test dispatcher to ensure collection started eagerly

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
MatthewTighe: Love this idea!

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
MatthewTighe: Let me know if this is better - updated some names as well to try and increase clarity

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/test/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometerTest.kt
```diff
@@ -0,0 +1,255 @@
...
+import org.junit.runner.RunWith
+import org.robolectric.Shadows
+import org.robolectric.annotation.Config
+import org.robolectric.shadows.SensorEventBuilder
+import org.robolectric.shadows.ShadowSensor
+import org.robolectric.shadows.ShadowSensorManager
+
+@RunWith(AndroidJUnit4::class)
+@Config(sdk = [28])
+class LifecycleAwareSensorManagerAccelerometerTest {
```
segunfamisa, line 30: 👌

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
segunfamisa: Thank you so much! It does help me. I now understand how the algo works.

segunfamisa (approved):
Thanks a lot. Changes look great to me. ✅

boek (approved):

PR #56, opened 2026-02-05
Bug 2011184 - check fragment is added before finding nav controller.

pollym:
We are seeing `IllegalStateException`s in the play store from calling `findNavController` when the fragment is not associated with a transaction (the parentFragmentManager is null) `isAdded()` is the recommended way to find out if this fragment is attached to an activity and  `findNavController()` is safe to call

NB this fix is based on static analysis only - plan is to keep an eye on the crash logs after it lands to confirm it works.

[running a try here](https://treeherder.mozilla.org/jobs?repo=try&revision=f74abab716cd9b4f9af2b1742294c4712f80644c) but it seems to have included every commit since dinosaurs roamed the earth, not sure what happened there


segunfamisa (approved):
Thank you! I think the speculative fix is okay, thank you.

segunfamisa (commented):
Do we need to do anything when the fragment is not added? Log an error with some state?

pollym:
> Do we need to do anything when the fragment is not added? Log an error with some state?

it's a good question, but i don't think so - suspect nobody would care about this error, and it would just create noise in the logs.
this branch would only happen if the fragment is being removed, like if the user launched it then quickly hit back or something, so it wouldn't be user-perceivable (if that is a word 😅)

PR #57, opened 2026-02-06
Bug 2015009 - truncate very long dialog messages.

pollym:
We are seeing ANRs in the play store from the compute line break algorithm when it is given extremely long words to process. Truncating these for the various prompts we have which accept any input.

[running a try here](https://treeherder.mozilla.org/jobs?repo=try&revision=5ccdd454cda92c53a620b1da48e218ec52348adc)

### DIFF
--- a/mobile/android/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AbstractPromptTextDialogFragment.kt
+++ b/mobile/android/android-components/components/feature/prompts/src/main/java/mozilla/components/feature/prompts/dialog/AbstractPromptTextDialogFragment.kt
@@ -14,6 +14,7 @@ import androidx.annotation.IdRes
 import androidx.core.view.isVisible
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import mozilla.components.feature.prompts.R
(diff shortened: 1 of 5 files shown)


pollym:
i have manually tested this locally with this web page
[alert.html](https://github.com/user-attachments/files/25129293/alert.html)

before:
https://github.com/user-attachments/assets/8cef60ed-ac2a-4698-a707-09f2c9d5aa9c

after:
https://github.com/user-attachments/assets/2fad8650-142a-4cbf-9a00-9adc48fd2dc7

shtrom:
[Just testing reviewer assignemnent]

pollym:
following feedback in review'n'brew, reduced the max char limit to 1200, because:
- it was still a little slow at 5000 chars, we were dropping frames
- the material dialog component truncates text longer than ~1275 chars itself anyway, so if you make it longer you can't see the ellipsis

moztcampbell:
+1 from me on the principle here of truncating with ellipsis. I'll leave someone who knows the UI code better to check the Kotlin (all looks fine to me though)

moyheen (approved):
Looks good, thanks for updating it!

PR #65, opened 2026-02-11
Bug 2016104 - Created s2s loading screen

fmasalha:


fmasalha:
The three gradient colors from the figma are hardcoded in GradientPalette:
  val gradientStart = Color(0xFF3A0F6E)   // dark purple
  val gradientMiddle = Color(0xFF7543E3)  // light purple
  val gradientEnd = Color(0xFFFF7638)     // orange

  Plus the shimmer colors in LoadingCanvas.kt:
  val SHIMMER_COLOR = Color(0xFF676280)
  val SHIMMER_TRANSPARENT = Color(0x00E6E0F5)
  
Im pretty sure that these colors dont exist in materialTheme or in acorncolors , or if they are going to be added by ux or if I have to add them or if hard codeing them is okay. Please let me know. Thanks!

fmasalha:
try attempt: https://treeherder.mozilla.org/jobs?revision=a00d89a4795ff99c2b20518f5a91e23172eacbe4&repo=try

MatthewTighe (changes requested):
Segun, Jeff, and I reviewed this during Pacific Review n Brew. Here are the major elements we would like to see changed:
1. let's pull out the shimmer text and handle that it in a separate follow-up
2. the `LoadingCanvas` should be something the consumer of the API can use agnostically, that is, we want to remove thngs like the drag handle and the text, and as much of the intermediate containers as possible. ideally, we would convert this to a Modifier extension function, so that it could be applied wherever we want. if that proves challenging, we could start by instead just exposing `GradientAnimationLayer`.

Our biggest goal here is getting an API we can apply to the existing bottom sheet, so we can get it onto a build and let UX start providing concrete feedback.

fmasalha:
try : https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=180789

mobile/android/android-components/components/feature/summarize/src/main/res/values/static_strings.xml
```diff
@@ -81,4 +81,6 @@
     <!-- Generic fallback message for the info error screen -->
     <string name="mozac_summarize_info_error_fallback_message">Unknown error</string>
 
+    <!-- Loading text displayed while a page summary is being generated -->
+    <string name="mozac_feature_summarize_loading_title" translatable="false">Creating summary\u2026</string>
```
boek: replace `\u2026` with `…`.

boek (approved):

mobile/android/android-components/components/feature/summarize/src/main/res/values/static_strings.xml
```diff
@@ -81,4 +81,6 @@
     <!-- Generic fallback message for the info error screen -->
     <string name="mozac_summarize_info_error_fallback_message">Unknown error</string>
 
+    <!-- Loading text displayed while a page summary is being generated -->
+    <string name="mozac_feature_summarize_loading_title" translatable="false">Creating summary…</string>
```
MatthewTighe, line 85: Let's just hardcode this string in Kotlin for now. The l10n team isn't set up to review string changes in Github. Any localized string changes need to go through Phabricator, and I would rather us just land this and handle strings later.

MatthewTighe (changes requested):

mobile/android/android-components/components/feature/summarize/src/main/res/values/static_strings.xml
boek: @MatthewTighe When we move all of the strings from `static_strings.xml` to `strings.xml` wouldn't they have a chance to review them then?

MatthewTighe: We resolved this offline, l10n has given the go-ahead to land static_strings patches here and handle the bulk review when we convert to `strings.xml`.

MatthewTighe (approved):

PR #70, opened 2026-02-20
Bug 2018332 - Added s2s settings screen

fmasalha:


fmasalha:
There are two things I want to highlight

1. The learn more button is a noop right now since Im not sure what url to send it to or if that sumo page exists.
2. I copied the settings names from seguns patch here https://phabricator.services.mozilla.com/D283755 and will wait for that to land than rebase this on main.

fmasalha:
try: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=181142

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/pagesummaries/PageSummariesSettingsMiddleware.kt
```diff
@@ -0,0 +1,40 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.pagesummaries
+
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.Store
+import org.mozilla.fenix.utils.Settings
+
+internal class PageSummariesSettingsMiddleware(
+    private val settings: Settings,
```
MatthewTighe: now that Segun's [patch](https://phabricator.services.mozilla.com/D283755#change-OWiUzyAJ1BiH) has landed, let's do a few things differently:
1. rebase this patch onto of that
2. rename `SummarizeFeatureDiscoverySettings` and its subtype to be `SummarizeFeatureSettings`.
3. add any properties needed for this patch to the interface
4. replace the `Settings` dependency in this middleware with the abstract `SummarizeFeatureSettings`

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/pagesummaries/PageSummariesSettingsAction.kt
```diff
@@ -0,0 +1,17 @@
...
+package org.mozilla.fenix.settings.pagesummaries
+
+import mozilla.components.lib.state.Action
+
+internal sealed interface PageSummariesSettingsAction : Action
+
+internal data class ChangeSummarizePagesPreference(val isChecked: Boolean) :
+    PageSummariesSettingsAction
+
+internal data class ChangeShakeToSummarizePreference(val isChecked: Boolean) :
```
MatthewTighe: Actions should reflect what has happened. Let's update these to `SummarizePagesPreferenceToggled` and `ShakeToSummarizePreferenceToggled`. Since the Store is the source of truth, we can also remove the `isChecked` parameter and manage the state in the reducer.

MatthewTighe (changes requested):
I think we should go ahead and rebase this over Segun's changes and make some adjustments accordingly. This is shaping up well though, thanks!

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/pagesummaries/PageSummariesSettingsFragment.kt
```diff
@@ -0,0 +1,61 @@
...
+                    val store = PageSummariesSettingsStore(
+                        initialState = PageSummariesSettingsState(
+                            summarizePagesEnabled = summarizeSettings.summarizePagesEnabled,
+                            shakeToSummarizeEnabled = summarizeSettings.shakeToSummarizeEnabled,
+                        ),
+                        reducer = ::pageSummariesSettingsReducer,
+                        middleware = listOf(
+                            PageSummariesSettingsMiddleware(
+                                settings = summarizeSettings,
+                                geminiNanoLlmProvider = GeminiNanoLlmProvider(),
```
fmasalha: I needed this to support choosing the correct string to display depending on if the local llm is available, not sure if this is sufficient or if we should wait for the llmpicker to be done to use it here.

fmasalha:
https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=181675

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/pagesummaries/PageSummariesSettingsAction.kt
```diff
@@ -0,0 +1,20 @@
...
+
+internal sealed interface PageSummariesSettingsAction : Action
+
+internal data object SummarizePagesPreferenceToggled : PageSummariesSettingsAction
+
+internal data object ShakeToSummarizePreferenceToggled : PageSummariesSettingsAction
+
+internal data object LearnMoreClicked : PageSummariesSettingsAction
+
+internal data object Init : PageSummariesSettingsAction
```
MatthewTighe: Let's make this a `ViewAppeared` action instead of `Init`. Actually, we may not even need it since we aren't supporting a multi-model design for V1

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/pagesummaries/PageSummariesSettingsReducer.kt
```diff
@@ -0,0 +1,31 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.settings.pagesummaries
+
+internal fun pageSummariesSettingsReducer(
+    state: PageSummariesSettingsState,
+    action: PageSummariesSettingsAction,
+) = when (action) {
+    SummarizePagesPreferenceToggled -> {
+        val newValue = !state.summarizePagesEnabled
```
MatthewTighe: nit:
```suggestion
        val newPagesEnabledValue = !state.summarizePagesEnabled
```

MatthewTighe (changes requested):
Let's update this so we can re-use the view in the bottom sheet when users click the settings gear https://www.figma.com/design/U6RLYEt44nGdFUluQ01LZW/%F0%9F%A4%9D-Shake-to-Summarize?node-id=15871-16176&m=dev.

We can host the view as a Composable through our summarization feature module, and then wire up the behavior separately for the bottom sheet and the settings separately. We will still probably want to define the settings as an interface that can be shared between the two views as well. The implementation of this interface could be wrapped into `SummarizeFeatureDiscoverySettings` as discussed in a previous comment

fmasalha:
try: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=182577

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/onboarding/FenixSummarizeFeatureSettings.kt
```diff
@@ -13,12 +13,20 @@ import org.mozilla.fenix.utils.Settings
  * Discovery settings for the summarize feature. It's a wrapper around [org.mozilla.fenix.utils.Settings] for easier
  * testing and narrowed API
  */
-class FenixSummarizeFeatureDiscoverySettings(
+class FenixSummarizeFeatureSettings(
     private val settings: Settings,
-) : SummarizeFeatureDiscoverySettings {
+) : SummarizeFeatureSettings {
     override val canShowFeature: Boolean
```
MatthewTighe: let's rename this to `isFeatureEnabled` while we're here

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/onboarding/SummarizeFeatureSettings.kt
```diff
@@ -5,13 +5,15 @@
...
 /**
- * Interface for managing the discovery of the summarization feature
+ * Interface for managing the discovery of the summarization feature.
  *
- * It helps to manage things like highlighting the menu item, and clearing it upon interaction
+ * Extends [SummarizeSettings] with discovery-specific behavior like highlighting the menu item
+ * and clearing it upon interaction.
  */
-interface SummarizeFeatureDiscoverySettings {
+interface SummarizeFeatureSettings : SummarizeSettings {
```
MatthewTighe: Now that I see this, I think we should avoid this multi-level hierarchy. Let's change this interface back to use the `Discovery` language, and then `FenixSummarizeFeatureSettings` can compose this and `SummarizationSettings`. I think we should also update the language in these existing types for clarity. 

I am kind of thinking of this as "configuration" maps to a read state and "settings" maps to a writable state. `FenixSummarizationFeatureConfiguration` is kind of an awkward name then because it does both, but I don't think we need to care enough to actually separate the use cases.  So:

1. `SummarizeFeatureDiscoverySettings` -> `SummarizationFeatureDiscoveryConfiguration`
2. `SummarizeSettings` -> `SummarizationFeatureSettings`
3. `class FenixSummarizationFeatureConfiguration : SummarizationFeatureDiscoveryConfiguration, SummarizationFeatureSettings`

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -175,6 +175,10 @@ class SettingsFragment : PreferenceFragmentCompat() {
             getPreferenceKey(R.string.pref_key_translation),
         )?.isVisible = FxNimbus.features.translations.value().globalSettingsEnabled &&
             components.core.store.state.translationEngine.isEngineSupported == true
+
+        findPreference<Preference>(
+            getPreferenceKey(R.string.pref_key_page_summaries),
+        )?.isVisible = Config.channel.isDebug
```
MatthewTighe: I think this should be `settings.shakeToSummarizeFeatureFlagEnabled`

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/settings/SummarizeSettingsMiddleware.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val settings: SummarizeSettings,
+    private val onLearnMoreClicked: () -> Unit,
+) : Middleware<SummarizeSettingsState, SummarizeSettingsAction> {
+
+    override fun invoke(
+        store: Store<SummarizeSettingsState, SummarizeSettingsAction>,
+        next: (SummarizeSettingsAction) -> Unit,
+        action: SummarizeSettingsAction,
+    ) {
+        next(action)
```
MatthewTighe, line 28: can we add a comment here:
`// allow the reducer to run first, so that we can accurately update our cache based on the updated state`

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationScreen.kt
```diff
@@ -93,30 +108,47 @@ private fun SummarizationScreen(
...
+                        middleware = listOf(
+                            SummarizeSettingsMiddleware(
+                                settings = summarizeSettings,
+                                onLearnMoreClicked = {},
+                            ),
+                        ),
+                    )
+                    SummarizeSettingsContent(store = settingsStore)
+                }
+            }
```
MatthewTighe, line 129: I don't think we should be doing all this setup here. We have a couple options:

1. Pass the settings store down instead of the settings. I think this is the worse but faster option, because I don't really think features should have two stores. 
2. Compose the settings "store" into our existing store. We could follow the precedent we set with the [CrashReportingAppMiddleware]. We would basically need to add matching definitions from the `SettingsStore` to the `SummarizationStore`, and then map from them to the types expected by the `SettingsMiddleware`. The ergonomics here are kinda gross, but this would be a valuable testbed for improvements. Probably makes sense to do the faster thing for now, but will you file a bug against the [S2S backlog](https://mozilla-hub.atlassian.net/browse/FXDROID-7398) if that ends up being your choice?

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationScreen.kt
```diff
@@ -184,9 +239,23 @@ private fun DragHandle(
...
+
+    **Why it matters:** Fans have long requested stories from this period, which explores the ancient conflict between the Jedi and the Sith at the height of their power.
+
+    **Key details:**
+    - Setting: The Old Republic, approximately 4,000 years before the Empire
+    - Director: Yet to be announced
+    - First film expected: 2028
+    - Will feature the original Sith Wars and the founding of the Jedi Order
+""".trimIndent()
+
```
MatthewTighe, line 201: YO what?!?!?! I have missed this news?!?!?!

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationScreen.kt
```diff
@@ -131,16 +163,37 @@ private fun SummarizationScreen(
 }
 
 @Composable
-private fun SummarizedContent(text: String, modifier: Modifier = Modifier) {
-    SelectionContainer(modifier = modifier) {
-        RichText(text = text)
+private fun SettingsAppBar(
```
MatthewTighe: Let's move this into the `Settings` package and make it internal

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationScreen.kt
```diff
@@ -154,8 +207,10 @@ private fun SummarizationScreenScaffold(
     ) {
         Column(
             modifier = Modifier
-                .padding(horizontal = AcornTheme.layout.space.static200)
-                .fillMaxWidth(),
+                .fillMaxWidth()
+                .thenConditional(
+                    Modifier.padding(horizontal = AcornTheme.layout.space.static200),
+                ) { applyContentPadding },
```
MatthewTighe: What is this conditional padding for? I am not seeing extra padding in the figma for the non-settings screens

MatthewTighe (changes requested):
Okay I think we're very close - mostly cosmetic changes outside of a few things, and one question on my end. Thanks for iterating on this!

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationScreen.kt
fmasalha: Its unfortunately all rumors and knowing the state of lucasfilm, it could get shelved or killed any day. 

Fingers crossed that it survives!

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationScreen.kt
fmasalha: This is actually to eliminate the padding that applies to the settings view in the sheet, which also applies to the other screens as it was developed.

`applyContentPadding = state !is SummarizationState.Settings,
`

If removed, the settings screen has way too much padding and gets squished. This is probably caused since the reusable composable has padding built into it rather than expected from the parent.

Not sure if thats fine or if i should remove the padding from the composable and apply at the parent level.

MatthewTighe: I think we should apply the padding at the parent level - the child shouldn't be making assumptions about how/where it is hosted

fmasalha: Sounds good

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationScreen.kt
fmasalha: Added this ticket for this, https://mozilla-hub.atlassian.net/browse/FXDROID-7669

I started working on option two but felt like it was going to continue to increase this patch size and I feel like its already too big.

fmasalha:
try: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=183156

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/browser/BrowserFragment.kt
```diff
@@ -179,7 +179,7 @@ class BrowserFragment : BaseBrowserFragment(), UserInteractionHandler {
     }
 
     private fun setupShakeDetection() {
-        if (!requireComponents.core.summarizeFeatureDiscoverySettings.canShowFeature) {
+        if (!requireComponents.core.summarizeFeatureSettings.canShowFeature) {
```
MatthewTighe: let's update this to be `&& shakeGestureEnabled`

MatthewTighe (approved):
Thanks for handling this! I think we should go ahead and land this to unblock future work, but please address the comment about using the shake gesture preference in the browser fragment before you land. Also, would you mind filing a follow-up to add learn more link functionality to all the places that need it?

PR #77, opened 2026-02-25
Bug 2019308 - Add fully qualified pacakge name to MLPA authorization request

segunfamisa:


segunfamisa:
try: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=181936

mobile/android/android-components/components/lib/llm-mlpa/src/main/java/mozilla/components/lib/llm/mlpa/MlpaTokenProvider.kt
```diff
@@ -71,17 +72,20 @@ fun interface MlpaTokenProvider {
          * @param authenticationService Service that verifies the integrity token
          * and exchanges it for an access token.
          * @param userIdProvider Supplies the current user's [UserId].
+         * @param packageName The package name of the app requesting verification.
          */
         fun mlpaIntegrityHandshake(
             integrityClient: IntegrityClient,
             authenticationService: AuthenticationService,
             userIdProvider: UserIdProvider,
+            packageName: PackageName,
```
segunfamisa, line 81: Wasn't sure if we needed a package name provider here. I also was not sure where this `mlpaIntegrityHandshake` would get called from, whether it's fenix or somewhere else.

If it's somewhere else, we might need to do a `PackageNameProvider` so that we are not forced to plumb down a `PackageName` param through all the call sites

mobile/android/android-components/components/lib/llm-mlpa/src/main/java/mozilla/components/lib/llm/mlpa/service/MlpaService.kt
```diff
@@ -89,11 +98,15 @@ fun interface AuthenticationService {
      *
      * @property userId The identifier of the user requesting verification.
      * @property integrityToken The integrity token obtained from the client.
+     * @property packageName The package name for the app requesting verification.
      */
     @Serializable
     data class Request(
         @SerialName("user_id") val userId: UserId,
-        @Serializable(with = IntegrityTokenSerializer::class) val integrityToken: IntegrityToken,
+        @SerialName("integrity_token")
```
segunfamisa, line 106: a drive-by fix that I made. It was not serializing correctly because we did not specify the serial name

boek (approved):
:shipit:

PR #86, opened 2026-03-02
Bug 2020490 - Creates a stable UUID to be used with Integrity and MLPA

boek:
[try](https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=182913)

### DIFF
--- a/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/ClientUUID.kt
+++ b/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/ClientUUID.kt
@@ -1,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package org.mozilla.fenix.components
+
+import android.content.Context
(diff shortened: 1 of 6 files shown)


mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/ClientUUID.kt
```diff
@@ -0,0 +1,47 @@
...
+) : UserIdProvider, RequestHashProvider {
+    private val uuid: String by lazy {
+        prefs.getString(KEY, null) ?: generateUUID().also {
+            prefs.edit { putString(KEY, it) }
+        }
+    }
+
+    override fun getUserId() = UserId(uuid)
+
+    override fun generateHash() = uuid
```
MatthewTighe, line 31: should we remove the `generateHash` API? doesn't seem semantically different to `getUserId` except it saves the caller an unwrap

MatthewTighe (approved):

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/ClientUUID.kt
boek: We're conforming to two different interfaces from two different packages so we need both of them for this one :)

PR #105, opened 2026-03-09
Bug 2021901 - Update FetchClientMlpaService headers

moztcampbell:
We are also required to send 'use-play-integriy: true' header, so update the request and tests.


fmasalha (approved):
LGTM Thanks for the patch!

PR #106, opened 2026-03-09
Bug 2022031 - Fixed frozen gradient issue

fmasalha:


boek (approved):

boek (changes requested):
We're going to add the withContext to the service

PR #113, opened 2026-03-10
Bug 2020529 - Created AI controls UI

fmasalha:


fmasalha:
I added the settings screen patch underneath just for now to add the nav button for it really. I will remove it once that lands separately.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsFragment.kt
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private fun openAiControlsSumoPage() {
+        val context = requireContext()
+        SupportUtils.launchSandboxCustomTab(
+            context = context,
+            url = SupportUtils.getSumoURLForTopic(
+                context = context,
+                topic = SupportUtils.SumoTopic.AI_CONTROLS,
+                useMobilePage = false,
```
fmasalha, line 41: Without this, the page linked in the figma doesnt get found

fmasalha: the "useMobilePage = false" ^

fmasalha:
try: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=184782

fmasalha:
FYI, all the toggles are essentially no-ops right now, all the toggles do not have the checkmark in spec'd in the figma as the reusable component doesn't offer that functionality rn. I made a ticket and added it to the s2s backlog for that.

fmasalha:
Note to self: Need to make the ai block all set the rest off instead of disabling the other toggles

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
```diff
@@ -53,20 +62,63 @@ fun TranslationSettings(
...
     onAutomaticTranslationClicked: () -> Unit,
     onNeverTranslationClicked: () -> Unit,
     onDownloadLanguageClicked: () -> Unit,
 ) {
     val showHeader = showAutomaticTranslations || showNeverTranslate || showDownloads
+    val effectiveEnabled = if (showMainToggle) translationsEnabled else true
 
     Surface {
         LazyColumn {
+            if (showMainToggle) {
```
MatthewTighe: I think we should either:
1. save the translations work for a follow-up.
2. move this conditional higher up in the view stack.

For 2, I think it would make sense to copy the existing TranslationsSettings into a new file/screen and name it something like `TranslationsSettingsWithAiControls`. Then, we can move the logic controlling the view hierarchy depending on the feature flag into the fragment. This would create a more robust logical split - it would be really easy to change or delete either version of the screen without worrying about impacting the other.

This would probably make the most sense to do as a separate patch (and remove the translations stuff from this one), but I'll that up to you. A case that makes me miss patch stacks 🥲

MatthewTighe (changes requested):

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
fmasalha: Ye sounds good, ill make a follow ticket and add it to the backlog. I wish we still had patch stacks aswell 😭 .

fmasalha:
https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=186096

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/translations/TranslationSettings.kt
fmasalha: https://mozilla-hub.atlassian.net/browse/FXDROID-7858

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+    voiceSearchEnabled: Boolean,
+    onBannerLearnMoreClick: () -> Unit,
+    onBlockAiToggle: (Boolean) -> Unit,
+    onSeeWhatsIncludedClick: () -> Unit,
+    onTranslationsToggle: (Boolean) -> Unit,
+    onMoreTranslationsClick: () -> Unit,
+    onPageSummariesToggle: (Boolean) -> Unit,
+    onMorePageSummariesClick: () -> Unit,
+    onVoiceSearchToggle: (Boolean) -> Unit,
+) {
```
MatthewTighe: I think we should go ahead and convert this to a `State`/`dispatcher` or even a `Store` to avoid the lengthy parameter list and prepare us for downstream changes

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+                    stringResource(R.string.app_name),
+                ),
+                maxDescriptionLines = Int.MAX_VALUE,
+                showSwitchAfter = true,
+                onClick = { newValue ->
+                    if (newValue) {
+                        showBlockDialog = true
+                    } else {
+                        onBlockAiToggle(false)
+                    }
```
MatthewTighe: I think we can wait to add business logic like this until a follow-up patch. This is generally true of this patch - we basically just end up doing the work twice when we migrate it to the correct design patterns.

mobile/android/android-components/components/ui/icons/src/main/res/drawable/mozac_ic_fox_ai_on_state.xml
```diff
@@ -0,0 +1,180 @@
+<!-- This Source Code Form is subject to the terms of the Mozilla Public
+   - License, v. 2.0. If a copy of the MPL was not distributed with this
+   - file, You can obtain one at https://mozilla.org/MPL/2.0/. -->
+
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
MatthewTighe, line 5: FYI, complex vector graphics like this can be slow to draw. We probably aren't at risk of a performance problem due to the simplicity of this screen, but in the future it can be good to try and use `.webp` illustrations that our UX partners can export for us. If we notice any jank on this screen, let's file a follow-up to fix

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsFragment.kt
```diff
@@ -0,0 +1,81 @@
...
+                    pageSummariesEnabled = pageSummariesEnabled,
+                    voiceSearchEnabled = voiceSearchEnabled,
+                    onBannerLearnMoreClick = { openAiControlsSumoPage() },
+                    onBlockAiToggle = {
+                        blockAiEnabled = it
+                        if (it) {
+                            translationsEnabled = false
+                            pageSummariesEnabled = false
+                            voiceSearchEnabled = false
+                        }
```
MatthewTighe: I don't think we need to include any business logic in this patch - we really just want the shell of the UI.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+            modifier = Modifier
+                .fillMaxSize()
+                .verticalScroll(rememberScrollState()),
+        ) {
+            AiChoiceBanner(onLearnMoreClick = onBannerLearnMoreClick)
+
+            SwitchListItem(
+                label = stringResource(R.string.ai_controls_block_ai_title),
+                checked = blockAiEnabled,
+                maxLabelLines = Int.MAX_VALUE,
```
MatthewTighe: I don't think we need more than 1 line for the label here

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+                    ),
+                    linkTextDecoration = TextDecoration.Underline,
+                    linkTextColor = MaterialTheme.colorScheme.tertiary,
+                )
+            }
+
+            Image(
+                painter = painterResource(iconsR.drawable.mozac_ic_fox_ai_on_state),
+                contentDescription = null,
+                modifier = Modifier.size(78.dp),
```
MatthewTighe: where did this 78 come from? I see the vector defines its own sizes, and the container uses 78x88. If we scaled to the container size though, we'd adjust the aspect ratio

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+private fun AiFeaturesSection(
+    enabled: Boolean,
+    translationsEnabled: Boolean,
+    pageSummariesEnabled: Boolean,
+    voiceSearchEnabled: Boolean,
+    onTranslationsToggle: (Boolean) -> Unit,
+    onMoreTranslationsClick: () -> Unit,
+    onPageSummariesToggle: (Boolean) -> Unit,
+    onMorePageSummariesClick: () -> Unit,
+    onVoiceSearchToggle: (Boolean) -> Unit,
```
MatthewTighe: Let's remove the callbacks and state from here too. We'll need to tie these into our registry at some point, so we'll need a dynamic way to list features from and their interactions anyway. We can just pass down or use empty lambdas as necessary

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+    SettingsLink(
+        text = stringResource(R.string.ai_controls_more_page_summary_settings),
+        onClick = onMorePageSummariesClick,
+    )
+
+    SwitchListItem(
+        label = stringResource(R.string.ai_controls_voice_search_title),
+        checked = voiceSearchEnabled,
+        enabled = enabled,
+        maxLabelLines = Int.MAX_VALUE,
```
MatthewTighe: We probably don't want max label lines here either

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+                top = 8.dp,
+                bottom = 16.dp,
+            ),
+    ) {
+        Row(
+            modifier = Modifier
+                .fillMaxWidth()
+                .clip(RoundedCornerShape(16.dp))
+                .background(FirefoxTheme.colors.layerAccentNonOpaque)
+                .padding(start = 16.dp, top = 12.dp, bottom = 12.dp),
```
MatthewTighe: Should the end padding match here for RTL locales?

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+                Spacer(modifier = Modifier.height(16.dp))
+
+                Text(
+                    text = buildAnnotatedString {
+                        withStyle(SpanStyle(fontWeight = FontWeight.Bold)) {
+                            append(whatBlocked)
+                        }
+                    },
+                    style = FirefoxTheme.typography.body2,
+                    color = MaterialTheme.colorScheme.onSurfaceVariant,
```
MatthewTighe: I think we can use `subtitle2` here instead of building an annotated string. At worst, we can `copy` the `body2` typography and adjust the weight

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+
+                BlockedFeatureItem(
+                    iconRes = iconsR.drawable.mozac_ic_lightning_24,
+                    label = stringResource(R.string.ai_controls_page_summaries_title),
+                )
+
+                BlockedFeatureItem(
+                    iconRes = iconsR.drawable.mozac_ic_microphone_24,
+                    label = stringResource(R.string.ai_controls_voice_search_title),
+                )
```
MatthewTighe: We'll probably need a way to populate these dynamically as well, but let's just leave these for now since they're simple.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+        dismissButton = {
+            TextButton(onClick = onCancel) {
+                Text(
+                    text = stringResource(R.string.ai_controls_block_dialog_cancel),
+                    color = MaterialTheme.colorScheme.primary,
+                )
+            }
+        },
+        confirmButton = {
+            TextButton(onClick = onBlock) {
```
MatthewTighe: Let's use our Acorn components for these text buttons: https://searchfox.org/firefox-main/source/mobile/android/android-components/components/compose/base/src/main/java/mozilla/components/compose/base/button/TextButton.kt#35

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AiControlsScreen.kt
```diff
@@ -0,0 +1,485 @@
...
+                    text = stringResource(R.string.ai_controls_block_dialog_block),
+                    color = MaterialTheme.colorScheme.error,
+                )
+            }
+        },
+    )
+}
+
+@Composable
+private fun BlockedFeatureItem(iconRes: Int, label: String) {
```
MatthewTighe: These can probably be Acorn `ListItem`: https://searchfox.org/firefox-main/source/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/compose/list/ListItem.kt

MatthewTighe (changes requested):
Went over this in Pacific Review n Brew today - things are coming together well but we didn't think there was any reason to include business logic in this patch. Let's try to keep it as dumb as possible. If you want to keep the SUMO link stuff in for now that's okay with me, but it's liable to change once we port this to a Store as well.

Also, generally speaking it's always helpful to include some screenshots when you submit patches with UI in them

mobile/android/android-components/components/ui/icons/src/main/res/drawable/mozac_ic_fox_ai_on_state.xml
fmasalha: Hmmm good to know. Thanks for the info.

fmasalha:
I was confused as to what the line was for business logic because I thought that we needed visibility for the ui elements like the dialog and extra banner etc, but now I understand what this patch was supposed to be. I went ahead and stripped everything that wasnt just the base ui and added previews for the dialog and banner.

MatthewTighe:
@fmasalha I think there was one comment that wasn't addressed here that I had an open question on:
https://github.com/mozilla-firefox/firefox/pull/113/changes#r2956273941

Other than that, could you post some screenshots of the preview please?

fmasalha:
> @fmasalha I think there was one comment that wasn't addressed here that I had an open question on: https://github.com/mozilla-firefox/firefox/pull/113/changes#r2956273941
> 
> Other than that, could you post some screenshots of the preview please?

I dont see that comment on the conversation tab which I assumed would display all comments, thanks for reminding me.

I put that there as a placeholder and forgot to circle back to it, thanks for catching that.

image

image

image

MatthewTighe (approved):
thanks for posting the previews! this LGTM to me now

PR #115, opened 2026-03-11
Bug 2022566 - Uplifted s2s gesture flag default to nightly

fmasalha:


segunfamisa (approved):

segunfamisa:
note about the testing flag: the feature flag is still gating the feature, so this does not require explicit testing. This only enables the shake gesture by default.

PR #122, opened 2026-03-13
Bug 2023236 - Added error reporting to sentry for s2s errors

fmasalha:


mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationMiddleware.kt
```diff
@@ -14,11 +14,23 @@ import mozilla.components.feature.summarize.content.PageContentExtractor
...
 
+/**
+ * Interface for reporting a throwable.
+ */
+fun interface ErrorReporter {
+
+    /**
+     * Reports the given [throwable].
+     */
+    fun report(throwable: Throwable)
```
fmasalha: The interface @boek mentioned to add included a suspend on report, but submitCaughtException isnt suspend so im not sure if that is required so i dropped it. Let me know if im missing something. Thanks!

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationMiddleware.kt
```diff
@@ -14,11 +14,23 @@ import mozilla.components.feature.summarize.content.PageContentExtractor
 import mozilla.components.lib.state.Middleware
 import mozilla.components.lib.state.Store
 
+/**
+ * Interface for reporting a throwable.
+ */
+fun interface ErrorReporter {
```
fmasalha: Not sure if this is a good or bad location for this to exist.

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationMiddleware.kt
boek: The crash reporter interface should be a suspending function because right now it is just returning a job: https://searchfox.org/firefox-main/source/mobile/android/android-components/components/lib/crash/src/main/java/mozilla/components/lib/crash/CrashReporter.kt#271

Lets make our interface good so if/when we update the crash reporter we won't have to fix it here as well.

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationMiddleware.kt
```diff
@@ -58,6 +70,7 @@ class SummarizationMiddleware(
                     }
             },
             onFailure = {
+                errorReporter.report(it)
                 store.dispatch(SummarizationFailed(it))
```
boek, line 137: Since we're dispatching an action here, let's listen to this action in the middleware and call errorReporter.report(it) there. That way we can report all of the errors in one place instead of of sprinkling this anywhere there could be an error. For example. We may also need to dispatch this for `LlmAction.ReceivedResponse.Failure`.

mobile/android/android-components/components/feature/summarize/src/test/java/mozilla/components/feature/summarize/SummarizationStoreTest.kt
```diff
@@ -22,6 +22,9 @@ import kotlin.time.Duration.Companion.seconds
 
 class SummarizationStoreTest {
 
+    private val reportedErrors = mutableListOf<Throwable>()
```
boek, line 29: You'll want to clear this out in test in a before all, otherwise this will collect anything in every test and make them less deterministic. Another thing that you could do is `val noopReporter = ErrorReporter { } ` and use it everywhere that we don't care about it and then in the tests where we want to test the reporter have `val reports = mutableListOf<Throwable>()` and just pass `{ reports.add(it) }` as the `errorReporter`

boek (changes requested):
This is close! just a couple of small tweaks and then its gtg

fmasalha:
try: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=186094

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationMiddleware.kt
MatthewTighe: I would say it should go in its own file, either in the top level package `feature/summarize` or in a new error package `feature/summarize/error`

MatthewTighe (approved):
LGTM! Left a suggestion for your open question, but feel free to land after you've made a choice there.

fmasalha:
try: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=187531

PR #126, opened 2026-03-16
Bug 2023538 - Skip "shake to summarize" CFR when a custom tab is active

segunfamisa:


segunfamisa:
try: https://treeherder.mozilla.org/jobs?repo=try&revision=33a8131c2adfefbdc576920d878c259711002416

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bindings/SummarizeToolbarCFRBinding.kt
```diff
@@ -57,7 +57,7 @@ class SummarizeToolbarCFRBinding(
 ) : AbstractBinding<BrowserState>(browserStore, mainDispatcher) {
 
     override suspend fun onState(flow: Flow<BrowserState>) {
-        flow.filter { it.selectedTab?.isNormalTab() == true }
+        flow.filter { it.selectedTab?.isNormalTab() == true && it.customTabs.isEmpty() }
```
MatthewTighe: It's possible to open a custom tab and then manually navigate back to the normal browser activity. Would this then prevent the CFR from triggering? Probably enough of an edgecase that we don't really care either way since this will be short-lived code.

MatthewTighe (approved):

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bindings/SummarizeToolbarCFRBinding.kt
segunfamisa: That's a good point. 

I chose this approach because I had thought that we cannot have custom tabs being non-empty while the browser is being focused.

How else can we make sure this does not happen in a custom tab?

segunfamisa: @MatthewTighe : I tried something different, I only applied the binding in `BaseBrowserFragment` when it is certain that we don't have a custom tab session.

It turns out that since we are in the base browser fragment and not the browser fragment, the BrowserStore is not enough to tell us if the current visible tab is a custom tab or not. We have to rely on knowledge of the fragment or presence/absence of a custom tab session id

segunfamisa:
new try:  https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=186244

PR #132, opened 2026-03-18
Bug 2024167 - Added haptic feedback on summary in s2s

fmasalha:


fmasalha:
Please let me if the feedback intensity is too low, personally I thought it might be too subtle but im not sure what the intended intensity is.

It could also be a per device case where I feel its subtle and others dont.

MatthewTighe (approved):

PR #133, opened 2026-03-18
Bug 2024176 - Updated shake consent composable to match figma spec

fmasalha:


mobile/android/android-components/components/feature/summarize/src/main/res/values/static_strings.xml
```diff
@@ -16,13 +16,13 @@
     <string name="mozac_summarize_learn_more_link">Learn more</string>
 
     <!-- Title for the off-device shake summarization consent screen -->
```
MatthewTighe: let's update this comment with the content localization note from the figma https://www.figma.com/design/Nrvne5sPQAUzMX9ysN9NMl/Shake-to-Summarize?node-id=2001-9783&m=dev

mobile/android/android-components/components/feature/summarize/src/main/res/values/static_strings.xml
```diff
@@ -16,13 +16,13 @@
...
-    <string name="mozac_summarize_shake_consent_off_device_title">Summarize this page</string>
+    <string name="mozac_summarize_shake_consent_off_device_title">Summarize with a shake?</string>
 
     <!-- Message for the off-device shake summarization consent screen  -->
```
MatthewTighe: same here, and we'll need to inform localizers what the variable is for, for example https://searchfox.org/firefox-main/rev/c1e060fc9eaba2d2e5fcac1d4227e680b6163730/mobile/android/android-components/components/browser/engine-system/src/main/res/values-ab/strings.xml#10

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/ui/OffDeviceSummarizationContent.kt
```diff
@@ -15,14 +17,24 @@ import androidx.compose.runtime.Composable
...
 import androidx.compose.ui.tooling.preview.PreviewLightDark
+import androidx.compose.ui.unit.dp
 import mozilla.components.compose.base.button.FilledButton
 import mozilla.components.compose.base.button.OutlinedButton
 import mozilla.components.compose.base.theme.AcornTheme
+import mozilla.components.feature.summarize.LocalProductName
 import mozilla.components.feature.summarize.OffDeviceSummarizationShakeConsentAction
 import mozilla.components.feature.summarize.R
 
+private val CONTENT_MIN_HEIGHT = 232.dp
```
MatthewTighe: I don't think this is a typical pattern - if we are needing additional spacing we should use spacers or padding. I see there are some spacers in the design that might work

Image

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationFragment.kt
```diff
@@ -115,7 +115,7 @@ class SummarizationFragment : BottomSheetDialogFragment() {
 
         FirefoxTheme {
             SummarizationUi(
-                productName = getString(R.string.app_name),
+                productName = getString(R.string.firefox),
```
MatthewTighe: I'm not sure whether we want this change. Could you check in with Nicole whether this should be the release variant name or always "Firefox"? It would need to be true everywhere in the module, otherwise we will need a second variable here

MatthewTighe (changes requested):

MatthewTighe:
Can you attach a screenshot of the changes as well please? Even just the preview would be fine

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationFragment.kt
fmasalha: Sorry I should have left a comment as to why I changed that, all the mentions of the app name in the figma is firefox in the consents screens and download screen, but app_name would be firefox fenix, so I assumed it must have been a mistake. The only other places that the product name variable is used is in those other consent screens. Let me know if you still want me to reach out to Nicole to double check.

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/ui/OffDeviceSummarizationContent.kt
fmasalha: The reason why I did this is because the figma had the spacing listed as auto, which I didn't know how to decipher other than it should take max space between each other which only worked if I defined a min height instead of guessing what the actual padding should be. I will swap it with a spacer instead though.

fmasalha:
Note: the reason why it says 'Firefox Debug' is because its a preview and that is the default for when the actual string isnt resolved.

Before:
Screenshot 2026-03-19 at 8 13 59 AM

After:
Screenshot 2026-03-19 at 8 13 27 AM

Figma:
image

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/ui/OffDeviceSummarizationContent.kt
fmasalha: So I swapped it to a static600 spaced spacer and it still looks a little short if you refer to the screenshots below, but its not that big of a difference. Let me know.

MatthewTighe: I think this is fine for now - these values are device dependent because different phones will have different resolution to pixel mappings. We won't necessarily be able to match designs perfectly, and this seems close enough to wait for design review later.

MatthewTighe (approved):

PR #136, opened 2026-03-19
Bug 2022850 - add game state

pollym:
- Fix issue where food could spawn underneath fox body and not be eaten. Previously, food was only eaten if it collided with the head :)
- Add some resources and draw code for the fox head, body, tail and food

### DIFF
--- a/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/DrawFoxBody.kt
+++ b/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/DrawFoxBody.kt
@@ -32,6 +32,27 @@ fun DrawScope.drawBody(state: GameState, shouldersPath: Path, bottomPath: Path) 
     val foxBody = state.fox.drop(1).dropLast(1)
     val cornerRadiusPx = state.cellSize / 2
     val cornerRadius = CornerRadius(cornerRadiusPx, cornerRadiusPx)
+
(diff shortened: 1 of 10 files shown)


pollym:
[green try here](https://treeherder.mozilla.org/jobs?repo=try&revision=29a6f42469b0dd5e498e5e44399059eec8a00234)

mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/DrawFoxBody.kt
```diff
@@ -0,0 +1,121 @@
...
+                    brush = brush,
+                    topLeft = Offset(x * state.cellSize, y * state.cellSize),
+                    size = Size(state.cellSize, state.cellSize),
+                )
+            }
+
+            val bottom = foxBody.last()
+            bottomPath.apply {
+                reset()
+                addRoundRect(
```
lottiehope, line 102: this logic is the same as the above in the else {} could you maybe extract it so it's not repeated? 🤔

lottiehope (approved):

mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/DrawFoxBody.kt
pollym: so it is. i will tinker!

PR #139, opened 2026-03-20
Bug 2024027 streaming mlpa

boek:


mobile/android/android-components/components/ui/richtext/src/main/kotlin/mozilla/components/ui/richtext/RichText.kt
```diff
@@ -63,6 +65,41 @@ fun RichText(
     }
 }
```
segunfamisa: Let's update this to use the overloaded API like:


```suggestion
    RichText(
        document = document,
        modifier = modifier,
        typography = typography,
        colors = colors,
        uriHandler = uriHandler,
    )
}
```

PR #147, opened 2026-03-23
Bug 2022850 - fix collision bug.

pollym:
fix edge case where collision detection could be erroneously triggered if the fox forms a perfect circle and moves into the space where the tail used to be.

I had a `drop()` where i should have had a `dropLast()`, basically i was looking at the fox upside down

### DIFF
--- a/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/GameState.kt
+++ b/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/GameState.kt
@@ -77,7 +77,7 @@ data class GameState(
             RIGHT -> head.copy(x = head.x + 1)
         }
 
-        val collidedWithSelf = newHead in fox.drop(1)
+        val collidedWithSelf = newHead in fox.dropLast(1)
         val collidedWithEdge = !withinBounds(newHead)
         val collidedWithFood = newHead == food || food in fox
         val isGameOver = collidedWithSelf || collidedWithEdge
--- a/mobile/android/fenix/app/longfox/src/test/kotlin/org/mozilla/fenix/longfox/GameStateTest.kt
+++ b/mobile/android/fenix/app/longfox/src/test/kotlin/org/mozilla/fenix/longfox/GameStateTest.kt
@@ -154,6 +154,57 @@ class GameStateTest {
         assertTrue(state.moveFox().isGameOver)
     }
 
+    @Test
+    fun `game not over when fox head moves into space where tail used to be`() {
+        //before move (<- direction = left)
+        //
+        //⬛️1️⃣2️⃣3️⃣4️⃣5️⃣
+        //1️⃣⬛️⬛️⬛️⬛️⬛️
+        //2️⃣⬛️⬛️⬛️⬛️⬛️
+        //3️⃣⬛️⬛️⬛️⬛️⬛️
+        //4️⃣⬛️⬛️🟧🦊🟧
+        //5️⃣⬛️⬛️🟧⬛️🟧
+        //6️⃣⬛️⬛️🟧🟧🟧
+
+        //after move
+        //
+        //⬛️1️⃣2️⃣3️⃣4️⃣5️⃣
+        //1️⃣⬛️⬛️⬛️⬛️⬛️
+        //2️⃣⬛️⬛️⬛️⬛️⬛️
+        //3️⃣⬛️⬛️⬛️⬛️⬛️
+        //4️⃣⬛️⬛️🦊🟧🟧
+        //5️⃣⬛️⬛️🟧⬛️🟧
+        //6️⃣⬛️⬛️🟧🟧🟧
+
+       val initialState = state(
+            direction = Direction.LEFT,
+            fox = listOf(
+                GridPoint(4, 4),
+                GridPoint(5, 4),
+                GridPoint(5, 5),
+                GridPoint(5, 6),
+                GridPoint(4, 6),
+                GridPoint(3, 6),
+                GridPoint(3, 5),
+                GridPoint(3, 4),
+            ),
+        )
+        val expectedFoxStateAfterMove = listOf(
+            GridPoint(3, 4),
+            GridPoint(4, 4),
+            GridPoint(5, 4),
+            GridPoint(5, 5),
+            GridPoint(5, 6),
+            GridPoint(4, 6),
+            GridPoint(3, 6),
+            GridPoint(3, 5),
+        )
+        assertFalse(initialState.isGameOver)
+        val newState = initialState.moveFox()
+        assertEquals(expectedFoxStateAfterMove, newState.fox)
+        assertFalse(newState.isGameOver)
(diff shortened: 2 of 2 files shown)


mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/GameState.kt
```diff
@@ -77,7 +77,7 @@ data class GameState(
             RIGHT -> head.copy(x = head.x + 1)
         }
 
-        val collidedWithSelf = newHead in fox.drop(1)
+        val collidedWithSelf = newHead in fox.dropLast(1)
         val collidedWithEdge = !withinBounds(newHead)
         val collidedWithFood = newHead == food || food in fox
         val isGameOver = collidedWithSelf || collidedWithEdge
```
titooan, line 83:
```suggestion
        val collidedWithFood = newHead == food || food in fox
        val collidedWithSelf = newHead in if (collidedWithFood) fox else fox.dropLast(1)
        val collidedWithEdge = !withinBounds(newHead)
        val isGameOver = collidedWithSelf || collidedWithEdge
```

I think this would consider better the case where the SnakeFox is currently eating food, meaning its tail will remain at the same position and its head will collide with the tail (because it stayed in the same position)

I'll put the test that makes it clearer in a separate comment. What do you think?

mobile/android/fenix/app/longfox/src/test/kotlin/org/mozilla/fenix/longfox/GameStateTest.kt
```diff
@@ -154,6 +154,57 @@ class GameStateTest {
...
+            GridPoint(4, 6),
+            GridPoint(3, 6),
+            GridPoint(3, 5),
+        )
+        assertFalse(initialState.isGameOver)
+        val newState = initialState.moveFox()
+        assertEquals(expectedFoxStateAfterMove, newState.fox)
+        assertFalse(newState.isGameOver)
+    }
+
```
titooan, line 207:
```Kotlin
@Test
    fun `game over when fox head moves into tail space after eating food`() {
        //before previous move (<- direction = left)
        //
        //⬛️1️⃣2️⃣3️⃣4️⃣5️⃣
        //1️⃣⬛️⬛️⬛️⬛️⬛️
        //2️⃣⬛️⬛️⬛️⬛️⬛️
        //3️⃣⬛️⬛️🟧⬛️⬛️
        //4️⃣⬛️⬛️🟧🍎🦊
        //5️⃣⬛️⬛️🟧⬛️🟧
        //6️⃣⬛️⬛️🟧🟧🟧
        //before move (<- direction = left)
        //
        //⬛️1️⃣2️⃣3️⃣4️⃣5️⃣
        //1️⃣⬛️⬛️⬛️⬛️⬛️
        //2️⃣⬛️⬛️⬛️⬛️⬛️
        //3️⃣⬛️⬛️⬛️⬛️
        //4️⃣⬛️⬛️🟧🦊🟧
        //5️⃣⬛️⬛️🟧⬛️🟧
        //6️⃣⬛️⬛️🟧🟧🟧

        //after move
        //
        //⬛️1️⃣2️⃣3️⃣4️⃣5️⃣
        //1️⃣⬛️⬛️⬛️⬛️⬛️
        //2️⃣⬛️⬛️⬛️⬛️⬛️
        //3️⃣⬛️⬛️⬛️⬛️⬛️
        //4️⃣⬛️⬛️💥🟧🟧
        //5️⃣⬛️⬛️🟧⬛️🟧
        //6️⃣⬛️⬛️🟧🟧🟧

        val initialState = state(
            direction = Direction.LEFT,
            fox = listOf(
                GridPoint(4, 4),
                GridPoint(5, 4),
                GridPoint(5, 5),
                GridPoint(5, 6),
                GridPoint(4, 6),
                GridPoint(3, 6),
                GridPoint(3, 5),
                GridPoint(3, 4),
            ),
            food = GridPoint(4, 4),
        )

        assertFalse(initialState.isGameOver)
        assertTrue(initialState.moveFox().isGameOver)
    }
```

That would be the test showing the case I'm describing and proposing to fix

titooan (commented):
I think there's a case the current patch misses, I've tried to describe a fix in the comments.

mobile/android/fenix/app/longfox/src/test/kotlin/org/mozilla/fenix/longfox/GameStateTest.kt
pollym: thank you for writing all that! the comments make it nice and clear what situation you're considering :)
however, i think this scenario is slightly mistimed, because it requires two frame ticks. But actually, the decision of whether to move the tail happens at the same time as we move the head onto food (in `moveFox()`, if we have `collidedWithFood` we don't return a copy with a moved tail). So i think the sequence of events would look like:

```
        // frame 1 (<- direction = left)
        //
        //⬛️1️⃣2️⃣3️⃣4️⃣5️⃣
        //1️⃣⬛️⬛️⬛️⬛️⬛️
        //2️⃣⬛️⬛️⬛️⬛️⬛️
        //3️⃣⬛️⬛️🟧⬛️⬛️
        //4️⃣⬛️⬛️🟧🍎🦊
        //5️⃣⬛️⬛️🟧⬛️🟧
        //6️⃣⬛️⬛️🟧🟧🟧
        //
        // frame 2 (<- direction = left) 
        // NB tail stays where it was
        //
        //⬛️1️⃣2️⃣3️⃣4️⃣5️⃣
        //1️⃣⬛️⬛️⬛️⬛️⬛️
        //2️⃣⬛️⬛️⬛️⬛️⬛️
        //3️⃣⬛️⬛️🟧⬛️⬛️
        //4️⃣⬛️⬛️🟧🦊🟧
        //5️⃣⬛️⬛️🟧⬛️🟧
        //6️⃣⬛️⬛️🟧🟧🟧
        //
        // frame 3: game over
        //
        //⬛️1️⃣2️⃣3️⃣4️⃣5️⃣
        //1️⃣⬛️⬛️⬛️⬛️⬛️
        //2️⃣⬛️⬛️⬛️⬛️⬛️
        //3️⃣⬛️⬛️⬛️⬛️⬛️
        //4️⃣⬛️⬛️💥🟧🟧
        //5️⃣⬛️⬛️🟧⬛️🟧
        //6️⃣⬛️⬛️🟧🟧🟧
```

I can add a test to verify this scenario! I feel like it wouldn't need code changes but i could be wrong

mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/GameState.kt
pollym: thanks for the suggestion! i don't think we need this though, the test i added below passes without it :)

titooan (commented):
That makes total sense to me, I had misunderstood the timing of the tail growing. Thanks for the explanation!

titooan (approved):

PR #149, opened 2026-03-23
Bug 2024961: Add game screen.

pollym:
This wires together the canvas, the game state and the sound effects, and sizes the game grid appropriately for the device.

Hoisted the screen size calculation out of the canvas to the new `LongFoxGameScreen` to simplify a bit.

NB there's no intro screen, sound on off. etc yet, that is to follow. Also it's not plumbed in anywhere in the actual browser yet, but we're adding the `LongfoxFeature.start()` function in this PR to allow you to do that soon :)

### DIFF
--- a/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
+++ b/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
@@ -1,0 +1,119 @@
+/*
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+package org.mozilla.fenix.longfox
+
+import androidx.compose.foundation.background
+import androidx.compose.foundation.gestures.detectTapGestures
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.BoxWithConstraints
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.material3.MaterialTheme
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.DisposableEffect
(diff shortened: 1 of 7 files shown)


pollym:
[here is a try](https://treeherder.mozilla.org/jobs?repo=try&revision=14661f9b9284d3dc5f9448572c5b4cd3264689e5)

pollym:
removed a rogue string resource: these will have to be landed separately via phabricator

mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
```diff
@@ -0,0 +1,119 @@
...
+        }
+        LaunchedEffect(gameState.isGameOver) {
+            if (gameState.isGameOver) soundEffectsPlayer.playSound(R.raw.sadwobble)
+        }
+        // This is the main game loop:
+        // While the game is not over, wait a clock tick, move the fox and check for collisions.
+        // Play a sound effect if that seems appropriate.
+        LaunchedEffect(gameState) {
+            while (!gameState.isGameOver) {
+                delay(GAME_INTERVAL_TIME_MS)
```
moyheen, line 78: What's the delay here for? To space out the sounds?

moyheen (approved):
This looks good to me

azinovyev (approved):
Looks good to me! Excited to see this coming together 🦊

AndreyZIno (approved):
Approved with the wrong account. Whoops.

mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
pollym: thanks for looking :D
good q! it's basically the delay between "frames" of the game; without this, the fox would move too quickly for a human to react to!
I thought about calling it "frame interval" but really it's a bit long for a frame, and in future i would like to maybe animate between states, so it wouldn't really map to a visual frame in that case.

PR #151, opened 2026-03-23
Bug 2025503 - Refactor summarization eligibility check from lambda into SummarizationEligibilityChecker in MenuDialogMiddleware

segunfamisa:


segunfamisa:
memory leaks try: https://treeherder.mozilla.org/jobs?repo=try&revision=f778ece87dc4314cd5e095613e993db03437fb34

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/menu/middleware/MenuDialogMiddleware.kt
```diff
@@ -165,7 +170,7 @@ class MenuDialogMiddleware(
             enabled = summarizeMenuSettings.showMenuItem &&
                     isNormalTab &&
                     !isLoading &&
-                    evaluateEligibilityForSummarization(),
+                    isEligibleForSummarization,
```
MatthewTighe: I had been intentionally invoking this lazily in case any of the earlier checks caused the conditional to short-circuit. It allows us to potentially avoid the work of fetching metadata through Gecko. I don't think we have to do it that way, but wanted to call it out in case you think it's worth changing.
```suggestion
        val isEligibleForSummarization = { store.state.browserMenuState?.selectedTab.checkSummarizationEligibility() }

val summarizationState = SummarizationMenuState.Default.copy(
visible = summarizeMenuSettings.showMenuItem,
highlighted = summarizeMenuSettings.shouldHighlightMenuItem && isNormalTab,
overflowMenuHighlighted = summarizeMenuSettings.shouldHighlightOverflowMenuItem && isNormalTab,
showNewFeatureBadge = true,
enabled = summarizeMenuSettings.showMenuItem &&
isNormalTab &&
!isLoading &&
                    isEligibleForSummarization(),
```

MatthewTighe (approved):
LGTM, thanks for handling this! Had one suggestion that we can handle in a follow-up if you think it's worth it, otherwise I will plan to land this

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/menu/middleware/MenuDialogMiddleware.kt
segunfamisa: That's a good shout and a brilliant one. I'll do that

PR #150, opened 2026-03-23
Bug 2024027 - Add streaming support to LLM APIs

MatthewTighe:
Taking over for https://github.com/mozilla-firefox/firefox/pull/143 since Jeff is on PTO

PR #152, opened 2026-03-24
WIP WIP WIP: Telemetry

fmasalha:
I created this sheet to better organize what I needed to make since the other two sheets attached to the ticket were a little difficult to follow imo,

https://docs.google.com/spreadsheets/d/1bQP-Ey2O2fl8vVFyvjru-vWM8zD568gxzbZphj7d-2g/edit?usp=sharing

### DIFF
--- a/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationAction.kt
+++ b/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationAction.kt
@@ -6,6 +6,7 @@ package mozilla.components.feature.summarize
 
 import mozilla.components.concept.llm.Llm
 import mozilla.components.concept.llm.LlmProvider
+import mozilla.components.feature.summarize.content.PageMetadata
 import mozilla.components.lib.state.Action
 
 /**
@@ -16,6 +17,9 @@ interface SummarizationAction : Action
 /** The Summarization Screen View Appeared */
 data object ViewAppeared : SummarizationAction
 
+/** The Summarization Screen View was Dismissed */
+data object ViewDismissed : SummarizationAction
+
 /** The user tapped the settings cog. */
 data object SettingsClicked : SummarizationAction
 
@@ -25,9 +29,17 @@ data object SettingsBackClicked : SummarizationAction
 /** Shake Consent has been requested */
 data object ShakeConsentRequested : SummarizationAction
 
-internal sealed interface LlmProviderAction : SummarizationAction {
+/**
+ * Actions related to the lifecycle of the [LlmProvider].
+ */
+sealed interface LlmProviderAction : SummarizationAction {
+    /** The LLM provider failed to initialize. */
     data object ProviderFailed : LlmProviderAction
+
+    /** The LLM provider is not yet available. */
     data object ProviderUnavailable : LlmProviderAction
+
+    /** The LLM provider finished initializing with the given [llm]. */
     data class ProviderInitialized(val llm: Llm) : LlmProviderAction
 }
 
@@ -36,9 +48,20 @@ internal sealed interface LlmProviderAction : SummarizationAction {
  */
 data class SummarizationFailed(val throwable: Throwable) : SummarizationAction
 
-/** Initialize the Llm */
-internal sealed interface LlmAction : SummarizationAction {
+/** Actions related to LLM interactions during summarization. */
+sealed interface LlmAction : SummarizationAction {
+    /** The user requested summarization; carries the provider [info]. */
     data class SummarizationRequested(val info: LlmProvider.Info) : SummarizationAction
+
+    /** The LLM has been prompted with extracted page content. */
+    data class LlmPrompted(
+        val instructions: String,
+        val content: String,
+        val pageMetadata: PageMetadata?,
+        val llm: Llm,
+    ) : LlmAction
+
+    /** A response was received from the LLM. */
     data class ReceivedResponse(val response: Llm.Response) : LlmAction
 }
 
--- a/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationMiddleware.kt
+++ b/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationMiddleware.kt
@@ -110,6 +110,12 @@ class SummarizationMiddleware(
             is SummarizationFailed -> scope.launch {
                 errorReporter.report(action.throwable)
             }
+            is LlmAction.LlmPrompted -> scope.launch {
+                action.llm.prompt(Prompt("${action.instructions} ${action.content}"))
+                    .collect { response ->
+                        store.dispatch(LlmAction.ReceivedResponse(response))
+                    }
+            }
         }
 
         next(action)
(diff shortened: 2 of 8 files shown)


mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationAction.kt
```diff
@@ -36,9 +48,20 @@ internal sealed interface LlmProviderAction : SummarizationAction {
  */
 data class SummarizationFailed(val throwable: Throwable) : SummarizationAction
 
-/** Initialize the Llm */
-internal sealed interface LlmAction : SummarizationAction {
+/** Actions related to LLM interactions during summarization. */
+sealed interface LlmAction : SummarizationAction {
```
fmasalha, line 52: I removed the internal so I can use it in the telemetry which is located in fenix since that is where all the telemetry packages live.

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationMiddleware.kt
```diff
@@ -128,10 +134,7 @@ class SummarizationMiddleware(
         }
         pageContentExtractor.getPageContent().fold(
             onSuccess = { content ->
-                llm.prompt(Prompt("$instructions $content"))
```
fmasalha, line 131: I split this up into two actions to make getting the telemetry data easier in the teleMiddleware

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationTelemetryMiddleware.kt
```diff
@@ -0,0 +1,174 @@
...
+                    AiSummarize.SummarizationClosedExtra(
+                        model = sessionTelemetry.model,
+                    ),
+                )
+            }
+            ShakeConsentRequested -> {
+                AiSummarize.summarizationConsentDisplayed.record(AiSummarize.SummarizationConsentDisplayedExtra())
+            }
+
+//            is OnDeviceSummarizationShakeConsentAction.AllowClicked,
```
fmasalha, line 89: So the telemetry google sheets had the description for the consent telemetry being when its displayed and than in the data science wish list it contained an extra for whether it was accepted or not. Should I make this into two separate telemetry signals or report on just when they agree or deny or just when displayed?

mobile/android/fenix/app/metrics.yaml
```diff
@@ -15025,3 +15025,197 @@ toolbar:
...
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=2025523
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=2025523
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: never
+  summarization_time:
```
fmasalha, line 15162: I did not implement this one yet as it is equivalent to the time extra that was requested in the summ_complete telemetry. Not sure which one to prio or to just keep both.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationTelemetryMiddleware.kt
MatthewTighe: I think we can continue using a single probe here and include the `agreed` extra. However, `CancelClicked` is not enough of a signal for this - we should also treat the `ViewDismissed` case as a "did not agree" is we are currently on a consent page.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationTelemetryMiddleware.kt
```diff
@@ -0,0 +1,174 @@
...
+                SummarizationTrigger.MENU
+            }
+            sessionTelemetry = sessionTelemetry.copy(trigger = trigger)
+        }
+    }
+
+    private fun handleLlmPrompted(action: LlmAction.LlmPrompted) {
+        sessionTelemetry = sessionTelemetry.copy(
+            contentMetrics = ContentMetrics(
+                wordCount = action.content.trim().split(Regex("\\s+")).size,
```
MatthewTighe, line 128: This isn't an accurate word count for non-English languages, as I recently learned in my page metadata patch 😅. Thankfully, a word count is already included in the page metadata! We will just need to propagate it into the SummarizationMiddleware through the `PageExtractor` mapping.

mobile/android/fenix/app/metrics.yaml
```diff
@@ -15025,3 +15025,197 @@ toolbar:
...
+    data_sensitivity:
+      - interaction
+    notification_emails:
+      - android-probes@mozilla.com
+    expires: never
+  summarization_completed:
+    type: event
+    description: |
+      Recorded when the summarizer service returns the summary or errors out.
+    extra_keys:
```
MatthewTighe, line 15051: Let's include content language here as well. It's not listed on the sheet, but I can imagine it being useful and it's available in the metadata already.

mobile/android/fenix/app/metrics.yaml
MatthewTighe: They're both useful. We can query timing distributions differently, and probably more cheaply for broad scale data.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationTelemetryMiddleware.kt
```diff
@@ -0,0 +1,174 @@
...
+        AiSummarize.summarizationCompleted.record(
+            AiSummarize.SummarizationCompletedExtra(
+                connectionType = connectionType.toString(),
+                contentType = sessionTelemetry.contentMetrics?.contentType,
+                errorType = errorType,
+                lengthChars = sessionTelemetry.contentMetrics?.charCount,
+                lengthWords = sessionTelemetry.contentMetrics?.wordCount,
+                model = sessionTelemetry.model.toString(),
+                outcome = outcome,
+                time = (System.currentTimeMillis() - sessionTelemetry.startTimeMillis).toInt(),
```
MatthewTighe, line 170: Since we are going to track a timing distribution as well, we can use that to get the time here. Should be able to do something similar to https://searchfox.org/firefox-main/rev/e49de7d10506041f63f620dc80088f3fb37901dd/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/home/recentsyncedtabs/RecentSyncedTabFeature.kt#178

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationMiddleware.kt
MatthewTighe: This makes sense to me. This area will need a careful rebase over https://github.com/mozilla-firefox/firefox/pull/150, where some of these names and the flow have changed. Let's also change the name from `LlmAction.LlmPrompted` to `ContentAction.ContentExtracted`.

MatthewTighe (changes requested):
Nice, this is coming along very well. We can prepare to land it, and then if anything needs to change as we add error types or land other patches we can do so. 

I've added some comments to address. I think we'll need to close this and re-open it in Phabricator, because I don't think the tooling for the data steward review is incorporated into the GH pilot yet.

MatthewTighe:
Closing this in favor of Phab for data review https://phabricator.services.mozilla.com/D289882

PR #154, opened 2026-03-24
Bug 2026042 - handle page metadata requests on the main thread

MatthewTighe:


fmasalha (approved):
LGTM! Thanks for the fix.

PR #155, opened 2026-03-25
Bug 2024710 - Swaped model target in MLPA request

fmasalha:


segunfamisa (approved):

mobile/android/android-components/components/lib/llm-mlpa/src/main/java/mozilla/components/lib/llm/mlpa/service/MlpaService.kt
```diff
@@ -276,7 +220,7 @@ fun interface ChatService {
                  * Predefined model identifier for the Mistral Small model hosted via Vertex AI.
                  */
                 val mistral: ModelID
```
segunfamisa: we should consider renaming this as well, from mistral to `mozSummarization` or something.

boek: 👍

PR #156, opened 2026-03-25
Bug 2026256 - Fine-tune shake gesture detection

segunfamisa:


segunfamisa:
try: https://treeherder.mozilla.org/jobs?repo=try&landoCommitID=188165

mobile/android/android-components/components/lib/shake/src/test/java/mozilla/components/lib/shake/ShakeDetectionTest.kt
```diff
@@ -5,196 +5,194 @@
...
+        assertEquals(
+            "Expected no shake is detected since sensitivity is above 1.g",
+            true,
+            shakes.isEmpty(),
         )
-
-        val accelerometer = FakeAccelerometer(samples)
-        val shakes = accelerometer.detectShakes(ShakeSensitivity.Medium).take(1).toList()
-
-        assertEquals(1, shakes.size)
```
segunfamisa, line 100: I think a bunch of these tests no longer apply, and I think they had an error anyway. We are taking the first emission from the flow, and also asserting that we detected only one shake.

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
```diff
@@ -55,47 +106,60 @@ private data class ShakeState(
...
-            }
-
-        // Step 3: Check if we're in the cooldown period after a previous shake detection.
+        val isExpired =
+            detectionWindowStartNs == 0L || timestampNs - detectionWindowStartNs > detectionWindowNs
+        val newHits = if (isExpired) {
+            1
+        } else if (shakeDirectionChanged) {
+            hits + 1
+        } else {
```
segunfamisa: This is the main change here, we only want to count a hit, if it is "roughly" in the opposite direction of the previous one.

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
```diff
@@ -55,47 +106,60 @@ private data class ShakeState(
...
+        val magnitude = sqrt(
+            sample.xAccel * sample.xAccel +
+                sample.yAccel * sample.yAccel +
+                sample.zAccel * sample.zAccel,
+        )
+        val timestampNs = sample.timestampNs
+
+        // Step 1: Check if acceleration magnitude is below the threshold.
+        // If it is below, reset detection state and wait for stronger acceleration.
+        if (magnitude < sensitivity.threshold) return copy(hasReachedMinHits = false)
```
MatthewTighe, line 129: I'm not sure this comment seems accurate to me - I would expect that we would want to reset the number of hits counted as well as the shake direction as part of resetting the detection state. Not saying the algorithm is wrong, but could we clarify the comment as to why we're only resetting `hasReachedMinHits`?

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
```diff
@@ -4,32 +4,51 @@
...
+ * @param cooldownPeriodNs The time in nanoseconds after a shake is detected. Any movement or shake that happens
+ * during this period is ignored
+ * @param minHits The minimum number of hits/pulses/rapid motion (above the threshold
+ * defined by [sensitivity]) that we need to see to determine that a shake has happened.
  */
-fun Accelerometer.detectShakes(sensitivity: ShakeSensitivity = ShakeSensitivity.Medium): Flow<Unit> =
+fun Accelerometer.detectShakes(
+    sensitivity: ShakeSensitivity = Medium,
+    detectionWindow: Long = 350_000_000L,
+    cooldownPeriodNs: Long = 800_000_000L,
```
MatthewTighe, line 37: nit, can we stay consistent on including the units? 
```suggestion
    detectionWindowNs: Long = 350_000_000L,
    cooldownPeriodNs: Long = 800_000_000L,
```

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
```diff
@@ -34,15 +33,15 @@ class LifecycleAwareSensorManagerAccelerometer(
 ) : Accelerometer, SensorEventListener, DefaultLifecycleObserver {
 
     private val sensor: Sensor? by lazy {
-        sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER)
+        sensorManager.getDefaultSensor(Sensor.TYPE_LINEAR_ACCELERATION)
```
MatthewTighe, line 36: Are we sure that this is a sensor that's present on all device types? Wondering if we need to include a fallback behavior if it isn't available.

MatthewTighe (approved):
This LGTM! I had one question that I think is worth answering that you may already know the answer to about the sensor type used. Other than that, I am still finding the "medium" sensitivity to not pass the throw it on the bed test, but the "low" sensitivity does and seems to still be sensitive enough to trigger shakes intentionally.

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
segunfamisa: Good catch. Thank you

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
segunfamisa: Hm. Tbh, I'm not sure. This code is not new, I just flipped the `!aboveThreshold` condition that was there before, to make it easier to parse.

I think that we should let the detection window determine whether or not we should reset the hits. From my POV, it is not impossible that we get a strong enough hit, and a weaker one, and then a strong one again, within the detection window, and I think that should still count as a shake.

segunfamisa: but i see what you say about the comment. I will remove the "reset detection state"

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
segunfamisa: My impression yesterday was that it would always be present, since it is a software fusion sensor, but I read around a bit more today, and found out that it combines gyroscope and accelerometer, and it is not impossible that the device has an accelerometer but not the other.

So I have added a fallback logic, and also handled normalization to a linear acceleration value.

PR #157, opened 2026-03-26
Bug 2026427 - add errors codes and handle them appropriately

MatthewTighe:
Could have maybe handled this in a few separate bugs, but once I got into it I figured it might be easiest to just get them done. This patch does the following:

1. associates an error code with various errors for reporting in the UI and in telemetry
2. adds the "content too long" error screen and uses it correctly
3. shows the error codes in the info error screen (this doesn't have designs, but I think it's valuable 🙈)
4. handles LLM providers returning an unavailable state better, instead of spinning on the "summarizing" screen

image

### DIFF
--- a/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationReducer.kt
+++ b/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationReducer.kt
@@ -30,16 +30,25 @@ fun summarizationReducer(state: SummarizationState, action: SummarizationAction)
         is SummarizationState.Settings -> SummarizationState.Summarized(info = state.info, document = state.document)
         else -> state
     }
-    is SummarizationFailed -> SummarizationState.Error(SummarizationError.SummarizationFailed(action.throwable))
+    is SummarizationFailed -> SummarizationState.Error(SummarizationError.SummarizationFailed(action.exception))
+    is LlmProviderAction.ProviderFailed -> SummarizationState.Error(
+        SummarizationError.SummarizationFailed(action.exception),
+    )
     else -> state
 }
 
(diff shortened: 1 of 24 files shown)


mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationReducer.kt
```diff
@@ -30,16 +30,25 @@ fun summarizationReducer(state: SummarizationState, action: SummarizationAction)
...
 }
 
 internal fun SummarizationState.applyResponse(response: Llm.Response): SummarizationState {
     if (this !is SummarizationState.Summarizing) return this
     return when (response) {
-        is Llm.Response.Failure -> SummarizationState.Error(
-            SummarizationError.SummarizationFailed(IllegalStateException(response.reason)),
-        )
+        is Llm.Response.Failure -> {
+            val contentTooLongErrorCode = 1005
```
MatthewTighe, line 44: I do hate this, but I'm not sure it seems that much different from knowing the status codes from a network API. Open to suggestions, but at the end of the day the behavior here might just be worth the bit of hackiness.

segunfamisa (approved):
we discussed the possibility of co-locating the error codes, and handling them in the middleware, then resolving each error code to a string in the composable.

but that's not a blocking comment, and can be done as a follow-up

MatthewTighe:
> we discussed the possibility of co-locating the error codes, and handling them in the middleware, then resolving each error code to a string in the composable.
> 
> but that's not a blocking comment, and can be done as a follow-up

I played with this a little bit and am finding the same thing Jeff did - the ergonomics of our choice to use a `Response` type from the LLM are just awkward. I think it's worth a bigger refactoring rather than trying to massage it into the least bad shape now, so I filed https://mozilla-hub.atlassian.net/browse/FXDROID-7993

PR #162, opened 2026-03-27
Bug 2026991: Add padding for non-edge-to-edge DohSettingsFragment.kt

pollym:
Recently, we replaced the forced activity level window insets padding with having individual fragments opt into the padding by implementing SystemInsetsPaddedFragment. Looks like this fragment was missed out.


pollym:
[running a try here](https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod&landoCommitID=188787)

moyheen (approved):
Thanks for jumping on this quickly! 🙏🏾

PR #169, opened 2026-03-30
Bug 2027872 - Improves memory performance of shake detection

boek:
This patch tries to reduce the number of allocations required while detecting a shake by
1. Introducing a ring buffer for recycling samples.
2. Replaces map/filter on the cold flow with a new flow that collects and emits when needed.
3. Replaces MutableSharedFlow with a Channel.

### DIFF
--- a/mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
+++ b/mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
@@ -11,8 +11,9 @@ import android.hardware.SensorManager
 import androidx.lifecycle.DefaultLifecycleObserver
 import androidx.lifecycle.LifecycleOwner
 import kotlinx.coroutines.channels.BufferOverflow
+import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.receiveAsFlow
 import mozilla.components.concept.accelerometer.Accelerometer
 import mozilla.components.support.base.log.logger.Logger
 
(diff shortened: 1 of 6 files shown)


boek:
[try](https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod&landoCommitID=189198)

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
```diff
@@ -32,35 +33,24 @@ class LifecycleAwareSensorManagerAccelerometer(
...
+
     private val sensor: Sensor? by lazy {
         sensorManager.getDefaultSensor(Sensor.TYPE_LINEAR_ACCELERATION)
             ?: sensorManager.getDefaultSensor(Sensor.TYPE_ACCELEROMETER)
     }
 
-    private val _samples = MutableSharedFlow<Accelerometer.Sample>(
-        extraBufferCapacity = NUM_BUFFER_CAPACITY,
+    private val _samples = Channel<Accelerometer.Sample>(
+        capacity = NUM_BUFFER_CAPACITY,
```
MatthewTighe, line 44: I assume the `10` here will get hit by magic number - but do we want the capacity of the custom buffer and the channel to be the same? Or does the channel even need to be buffered if we're maintaining our own?

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/SensorSample.kt
```diff
@@ -0,0 +1,56 @@
...
+     * Returns the next slot in the ring buffer, populated with the given values.
+     */
+    fun request(
+        xAccel: Float,
+        yAccel: Float,
+        zAccel: Float,
+        timestampNs: Long,
+    ): SensorSample {
+        val index = index.also {
+            index = (index + 1) % samples.size
```
MatthewTighe: nit: can we disambiguate the local variable from the class property? `updatedIndex` or `newIndex` or something?

MatthewTighe: I do love this modulo calculation

mobile/android/android-components/components/lib/shake/src/main/java/mozilla/components/lib/shake/ShakeDetection.kt
```diff
@@ -103,19 +104,19 @@ private enum class ShakeDirection {
...
-    val hits: Int = 0,
-    val detectionWindowStartNs: Long = 0L,
-    val lastShakeNs: Long = 0L,
-    val lastShakeDirection: ShakeDirection = ShakeDirection.Unknown,
-    val hasReachedMinHits: Boolean = false,
+    var hits: Int = 0,
+    var detectionWindowStartNs: Long = 0L,
+    var lastShakeNs: Long = 0L,
+    var lastShakeDirection: ShakeDirection = ShakeDirection.Unknown,
+    var hasReachedMinHits: Boolean = false,
```
MatthewTighe, line 111: curse you JVM 💔

MatthewTighe (approved):
Few small suggestions but nothing blocking. Awesome work! This is a cool resolution

mobile/android/android-components/components/lib/accelerometer-sensormanager/src/main/java/mozilla/components/lib/accelerometer/sensormanager/LifecycleAwareSensorManagerAccelerometer.kt
boek: `10` wasn't caught by magic number 🤷, I can make it a constant. The SampleBuffer is the `RingBuffer` not the buffer for the channel. We still need both.

PR #171, opened 2026-03-31
Bug 2028126: Optimise draw code.

pollym:
Move all the calculation out of the draw method, instead passing in a data object with all the values it needs.

Also reuse the expensive Path object rather than creating a new one each time.


pollym:
[here is a try](https://treeherder.mozilla.org/jobs?repo=try&revision=acd78e1f2a60d95d19dfad6248fc664ba3eb20b5)

lottiehope (approved):

PR #173, opened 2026-03-31
Bug 2028237 - Cleans up summarization middleware

boek:
This patch does a few things:

1. Removes the `Llm.Response` abstraction and leans on `Flow`.
2. Cleans up the content extracted action for the telemetry middleware
3. Pulls the business logic for observing the `Llm` and `LlmProvider` into extensions to clean up the `SummarizationMiddleware`

### DIFF
--- a/mobile/android/android-components/components/concept/llm/src/main/java/mozilla/components/concept/llm/Llm.kt
+++ b/mobile/android/android-components/components/concept/llm/src/main/java/mozilla/components/concept/llm/Llm.kt
@@ -23,10 +23,12 @@ value class ErrorCode(val value: Int)
  */
 interface Llm {
     /**
-     * A prompt request delivered to the LLM for inference, which will stream a series
-     * of [Response]s as they are made available.
+     * A prompt request delivered to the LLM for inference.
+     *
+     * @param prompt a [Prompt] that will be sent to the [Llm].
+     * @return a [Flow] of [String] of the response from the [Llm].
      */
-    suspend fun prompt(prompt: Prompt): Flow<Response>
+    suspend fun prompt(prompt: Prompt): Flow<String>
 
     /**
      * An exception thrown by an LLM, equipped with an [ErrorCode] to differentiate
@@ -48,33 +50,5 @@ interface Llm {
                 errorCode = ErrorCode(0),
             )
         }
-    }
-
-    /**
-     * A response from prompting a LLM.
-     */
-    sealed class Response {
-
-        /**
-         * A successful response from the LLM has occurred. This may include partial data,
-         * or be an indication that the reply has completed.
-         */
-        sealed class Success : Response() {
-            /**
-             * A (potentially) partial reply from the LLM. This may be a complete reply if
-             * it is short or the underlying implementation does not stream responses.
-             */
-            data class ReplyPart(val value: String) : Success()
-
-            /**
-             * An indication that the reply from the LLM is finishes.
-             */
-            data object ReplyFinished : Success()
-        }
-
-        /**
-         * A failure response from a LLM.
-         */
-        data class Failure(val exception: Llm.Exception) : Response()
     }
 }
--- a/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationAction.kt
+++ b/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationAction.kt
@@ -46,7 +46,7 @@ sealed interface LlmProviderAction : SummarizationAction {
 /**
  * There was a failure in summarizing content from the current page.
  */
-data class SummarizationFailed(val exception: Llm.Exception) : SummarizationAction
+data class SummarizationFailed(val throwable: Throwable) : SummarizationAction
 
 /**
  * We've requested a response from a Llm.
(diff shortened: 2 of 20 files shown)


boek:
[try](https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod&landoCommitID=189518)

segunfamisa:
@boek I think try is not happy. A couple of test failures.

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/ext/Llm.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    val parser = Parser()
+    val responseBuilder = StringBuilder()
+
+    prompt(Prompt("${pageMetadata.systemPrompt} $content"))
+        .map { responseBuilder.append(it) }
+        .sampledMap { parser.parse(it.toString()) }
+        .flowOn(dispatcher)
+        .collect { emit(ReceivedParsedDocument(it)) }
+}
```
segunfamisa: This block has enough "middleware-y" stuff, for me to think we should co-locate it in the middleware.

To me, it seems more relevant to the middleware (since the result of this operation is a middleware action), than it is an extension on LLM.

I would say we could put this functionality in the middleware directly....if someone is reading the code and they come across `llm.summarizeContent(...)` they would not be able to tell that this also emits content extracted & received parsed documents....it's even harder to discover if they are in separate files....and it's also an extension on "LLM" abstraction.

I don't hold this strongly enough to block the PR, but I wanted to describe how I perceived it, while reading the PR

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationAction.kt
```diff
@@ -46,31 +46,29 @@ sealed interface LlmProviderAction : SummarizationAction {
...
 /**
  * Page content has been extracted and is ready to be sent to the LLM.
  */
 data class ContentExtracted(
-    val instructions: String,
-    val content: String,
-    val pageMetadata: PageMetadata?,
-    val llm: Llm,
+    val pageMetadata: PageMetadata,
+    val charCount: Int,
```
segunfamisa: Probably something unrelated, but I find it interesting that we are more interested in char count, than in word count :/

mobile/android/android-components/components/concept/llm/src/main/java/mozilla/components/concept/llm/Llm.kt
```diff
@@ -49,32 +51,4 @@ interface Llm {
...
-             * An indication that the reply from the LLM is finishes.
-             */
-            data object ReplyFinished : Success()
-        }
-
-        /**
-         * A failure response from a LLM.
-         */
-        data class Failure(val exception: Llm.Exception) : Response()
-    }
```
segunfamisa, line 79: Yeah I like this.

The APIs were initially designed without streaming in mind...and then we started to retrofit streaming (flow) so, that kind of should have killed the response type. So I'm happy this is happening. Thank you!

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/ext/Llm.kt
```diff
@@ -0,0 +1,129 @@
...
+
+internal fun Llm.summarizeContent(
+    pageContentExtractor: PageContentExtractor,
+    pageMetadataExtractor: PageMetadataExtractor,
+    dispatcher: CoroutineDispatcher,
+) = flow {
+    val pageMetadata = pageMetadataExtractor.getPageMetadata()
+        .getOrDefault(PageMetadata())
+
+    val content = pageContentExtractor.getPageContent().getOrThrow()
```
segunfamisa: I think that, either here, or in the `PageContentExtractor` and `PageMetadata`, we should ensure that these run in the `Main` dispatcher context.

GeckoView errors otherwise, and I believe our APIs should accommodate developer mistakes (I made a similar mistake last week, by switching to a CPU-bound dispatcher to contact geckoview)

mobile/android/android-components/components/lib/llm-mlpa/src/main/java/mozilla/components/lib/llm/mlpa/MlpaLlm.kt
```diff
@@ -20,22 +19,9 @@ internal class MlpaLlm(
...
-                    ?.let {
-                        val exception = it as? Llm.Exception
-                            ?: Llm.Exception(it.message ?: "unknown chat error", unknownLlmErrorCode)
-                        Llm.Response.Failure(exception)
-                    }
-                    ?: Llm.Response.Success.ReplyFinished
-                emit(action)
-            }
-            .collect { emit(Llm.Response.Success.ReplyPart(it)) }
+    override suspend fun prompt(prompt: Prompt): Flow<String> = flow {
```
segunfamisa: I feel like instead of collecting and re-emitting, now that the API is simpler, we could call the chat service directly and return the flow for continuation....something like:


```suggestion
    override suspend fun prompt(prompt: Prompt): Flow<String> = 
        chatService.completion(authorizationToken, prompt.asRequest)    
    
```

segunfamisa (approved):
Thanks for this PR.

I don't think any of the comments I made are serious enough to block the PR. 

If you agree with some of the points, I think we can fix the broken tests from that try, and land this. 
Then we can file a follow-up in our backlog so we don't lose track of it.

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationAction.kt
boek: I thought the same thing but I looked at the doc from data science and they're requesting both.

boek:
[try](https://treeherder.mozilla.org/jobs?repo=try&revision=98bf2140ab2a6d252e1aa573396831bbc1e7a2ca)

segunfamisa (approved):

boek:
[try](https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod&landoCommitID=189798)

boek:
[try](https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod&landoCommitID=189820)

PR #174, opened 2026-04-01
Bug 2020527 - Add page summaries to AI Controls

MatthewTighe:


segunfamisa (approved):

segunfamisa:
Do we have a try for this? We could do rebase + try

fmasalha (commented):
I think we might need to tie the block toggle visibility to the feature flag if we are doing a locale rollout so that it doesnt show up if they dont have access to page summaries.

Not sure if that would be controlled by the feature flag or something else and should be done in a follow up.

MatthewTighe:
try https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod&landoCommitID=189778

PR #175, opened 2026-04-01
Bug 2028367: Add back button.

pollym:
- also darken the bricks so it has enough contrast to be visible
- extract some colours
- move the score to the opposite corner
- also leave the score in place on the new game screen if it is more than zero, so people can admire their last score

image

[try is here](https://treeherder.mozilla.org/jobs?repo=try&revision=aa02d5e0b0fc925de3ce6dbad49aed1d4205b068)

### DIFF
--- a/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
+++ b/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
@@ -6,16 +6,23 @@
 
 package org.mozilla.fenix.longfox
 
+import androidx.activity.compose.LocalOnBackPressedDispatcherOwner
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.clickable
 import androidx.compose.foundation.gestures.detectTapGestures
+import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.padding
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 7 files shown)


mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
```diff
@@ -130,8 +139,25 @@ fun LongFoxGameScreen() {
...
+            Image(
+                painter = painterResource(id = R.drawable.outline_arrow_back_24),
+                modifier = Modifier
+                    .padding(top = 12.dp, bottom = 12.dp, end = 12.dp)
+                    .clickable {
+                        onBackPressedDispatcher?.onBackPressed()
+                    },
+                contentDescription = "back" // TODO add string
+            )
+            if (gameState.score > 0) {
```
AndreyZIno, line 158: Do we want the score to only show when we have more than 0 points? So if a user lost before getting any points we wouldn't show 0?

Mainly just curious as to why we don't want to show 0 points.

AndreyZIno (commented):
This looks good! I think there might be some linting errors, for example commas for the Modifier chains. Also just making sure don't care about adding the strings to strings.xml at this time right?

AndreyZIno (approved):

mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
pollym: good q! yeah, i guess i just thought there's not much of a reason to show 0 points. it's not very motivating 😅 
also might be a bit weird to show 0 points the first time we come to the screen, if this is supposed to show the last score - this is an easy way to deal with that.

pollym: for the strings, they have to go through phabricator so there is a [separate patch here](https://phabricator.services.mozilla.com/D291492) :)

PR #189, opened 2026-04-08
Bug 2029590 - Removes the NEW label from summarize page when disabled

boek:
The badge color is hard coded deeper in the compose tree so I opted to remove it if summarization is disabled.

image


MatthewTighe (approved):
LGTM

PR #192, opened 2026-04-08
Bug 2029610 - Changes the color of the Firefox logo on the summarization loading screen

boek:
image


fmasalha:
https://www.figma.com/design/Nrvne5sPQAUzMX9ysN9NMl/Shake-to-Summarize?node-id=2001-9533&m=dev

image


Are they using an older figma?

This was the correct state in the handoff.

fmasalha:
I see now another dark mode in the same figma

image

When i was making this screen, nicole tagged the first screenshot i sent as the darkmode, not sure if this was changed to be the dark logo after it landed or if the dark logo was included in the handoff by accicdent.

boek:
I believe that it changed after we initially built it.

MatthewTighe (approved):

PR #197, opened 2026-04-09
Bug 2030463 - Fixed ai controls voice search truncate

fmasalha:


fmasalha:
The reusable composable we use here maps the max description lines parameter to a text composable which defaults to max lines anyways so I figured its probably fine to use here

@ Composable
fun Text(
    ….
    maxLines: Int = Int.MAX_VALUE,
    ....

boek (approved):

PR #209, opened 2026-04-13
Bug 2030740 - Update default summarization prompt

MatthewTighe:


boek (approved):

boek:
Tests use the default instructions as part of their assertions

PR #215, opened 2026-04-15
Bug 2024810 - Fixed last saved folder cache updating

fmasalha:


mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksMiddleware.kt
```diff
@@ -243,9 +243,7 @@ internal class BookmarksMiddleware(
                     }
 
                     preReductionState.bookmarksEditBookmarkState != null -> {
-                        if (!getNavController().popBackStack()) {
-                            exitBookmarks()
-                        }
+                        val popped = getNavController().popBackStack()
                         lifecycleScope.launch {
```
fmasalha, line 247: The scope was changed to be a cancel able scope in February which made the following code not run since exitBookmarks would destroy the fragment which cancels the lifecycleScope causing the last saved pref not to get updated.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMiddleware.kt
```diff
@@ -491,8 +491,17 @@ class BrowserToolbarMiddleware(
 
                 selectedTab?.let {
                     scope.launch(ioDispatcher) {
-                        val parentGuid = settings.lastSavedFolderCache.getGuid() ?: BookmarkRoot.Mobile.id
-                        val parentNode = bookmarksStorage.getBookmark(parentGuid).getOrNull()
+                        val targetParentFolderId =
```
fmasalha, line 494: You can add a bookmark star to the toolbar which didnt have the previously added safeguard to the menudialog.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMiddleware.kt
```diff
@@ -491,8 +491,17 @@ class BrowserToolbarMiddleware(
...
+                            settings.lastSavedFolderCache.getGuid() ?: BookmarkRoot.Mobile.id
+
+                        val parentNode = bookmarksStorage.getBookmark(targetParentFolderId).getOrNull()
+                            ?: bookmarksStorage.getBookmark(BookmarkRoot.Mobile.id).getOrNull()
+                        val parentGuid = parentNode?.guid ?: BookmarkRoot.Mobile.id
+
+                        if (targetParentFolderId != parentGuid) {
+                            settings.lastSavedFolderCache.setGuid(null)
+                        }
+
```
segunfamisa, line 504: To be honest, I think that we should fix this once and for-all, and make our `AddBookmarksUseCase` do all of this dance.

This is pretty much duplicated code from the `MenuDialogMiddleware`, doing the same thing.

https://github.com/mozilla-firefox/firefox/blob/662d18616bbb2da4c5a8758da50fc9d591d18c19/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/menu/middleware/MenuDialogMiddleware.kt#L265-L280

We [discussed this in January](https://github.com/mozilla-firefox/firefox/pull/33#issuecomment-3783446096) and we all kinda agreed that we need to unify this functionality. I think this is a good opportunity to do so.

I also see that you clear the last saved folder cache if for some reason, the resolved parentGuid is not matching the last saved parent folder. That's a good improvement

So, instead of duplicating this logic, we could consider the following:
1. Move `LastSavedFolderCache` somewhere in our bookmarks package or something
2. Add it as a parameter to our [AddBookmarksUsecase](https://github.com/mozilla-firefox/firefox/blob/06746d2778f78a4965ae0d094642d5094364513f/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/bookmarks/BookmarksUseCase.kt#L28)
3. Move this logic into the use-case.
4. Remove similar duplication that we do in various places
5. Update tests.

I will say that we could land this to do this properly now, but that's what we said the last time 😂 

Maybe @MatthewTighe  & @boek  can weight in here.

segunfamisa (approved):
Thanks for this PR. I added a comment about something I think we should fix once-and for all, but also okay with having this as a follow-up

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/components/toolbar/BrowserToolbarMiddleware.kt
fmasalha: I agree, I think this is a larger amount of work so I filed a follow up for this here https://bugzilla.mozilla.org/show_bug.cgi?id=2032600

fmasalha:
testing-tag explanation 

The follow up work will streamline this approach to make it testable 

https://bugzilla.mozilla.org/show_bug.cgi?id=2032600

PR #220, opened 2026-04-16
Bug 2032587 - Added UI for importing bookmarks

fmasalha:


mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksAction.kt
```diff
@@ -158,3 +158,8 @@ internal sealed class SnackbarAction : BookmarksAction {
     data object Dismissed : SnackbarAction()
     data object SelectFolderFailed : SnackbarAction()
 }
+
+internal data object RootMenuClicked : BookmarksAction
```
segunfamisa, line 162: What's root menu in this case? I think "overflow menu" is probably more common so maybe "RootOverflowMenu...." is best here.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksScreen.kt
```diff
@@ -285,6 +288,17 @@ private fun BookmarksList(
...
+        contract = ActivityResultContracts.GetContent(),
+    ) { _ -> }
+
+    LaunchedEffect(state.launchFilePicker) {
+        if (state.launchFilePicker) {
+            filePickerLauncher.launch("text/html")
+            store.dispatch(FilePickerDismissed)
+        }
+    }
+
```
segunfamisa, line 301: We can make this into a `FilePicker.kt` file and move it out of `BookmarksScreen.kt`. This file is already way too big.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksScreen.kt
```diff
@@ -863,6 +877,18 @@ private fun BookmarksListTopBarActionsNoSelection(
                 tint = MaterialTheme.colorScheme.onSurface,
             )
         }
+    } else if (LocalContext.current.settings().importBookmarksFeatureFlagEnabled) {
```
segunfamisa, line 880: Let's put something like `showBookmarksImport` in the state. We should not be accessing settings or data sources directly in the UI.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/bookmarks/BookmarksScreen.kt
```diff
@@ -1307,43 +1337,84 @@ private fun EmptyList(
...
-                text = stringResource(state.descriptionId()),
-                color = MaterialTheme.colorScheme.onSurfaceVariant,
-                style = FirefoxTheme.typography.body2,
-                textAlign = TextAlign.Center,
-            )
+    Spacer(modifier = Modifier.height(FirefoxTheme.layout.space.static300))
 
-            if (state is EmptyListState.NotAuthenticated) {
-                Spacer(modifier = Modifier.height(FirefoxTheme.layout.space.static300))
+    if (LocalContext.current.settings().importBookmarksFeatureFlagEnabled) {
```
segunfamisa, line 1370: Same here. Let's put this in the state instead.

mobile/android/fenix/app/src/main/res/values/strings.xml
```diff
@@ -1973,6 +1973,8 @@
     <string name="bookmark_sort_menu_a_to_z">Sort by A to Z</string>
     <!-- Label for button to sort bookmarks in reverse alphabetical order in the bookmark sorting overflow menu. -->
     <string name="bookmark_sort_menu_z_to_a">Sort by Z to A</string>
+    <!-- Label for button to import bookmarks from file. -->
+    <string name="bookmark_import_menu_button">Import from file</string>
```
segunfamisa, line 1977: This PR should go on phabricator because it touches strings.

segunfamisa (changes requested):
I think we're almost there, but there are a couple of changes I would like to recommend.

I also think we should move this over to phabricator, since it touches strings.xml files. Or is that no longer applicable?

mobile/android/fenix/app/src/main/res/values/strings.xml
fmasalha: Ahhh I forgot lol. I wonder how much longer we are gonna forced to use phab for patches that touch strings.

segunfamisa: LOL. I hope not for too long!

PR #222, opened 2026-04-17
Bug 2032634 - Adds interface for inserting a tree of bookmarks

boek:


mobile/android/android-components/components/concept/storage-bookmarks/src/main/java/mozilla/components/concept/storage/bookmarks/BookmarksStorage.kt
```diff
@@ -0,0 +1,68 @@
...
+    ) : InsertableBookmarkNode
+
+    /**
+     * A bookmark separator.
+     */
+    object Separator : InsertableBookmarkNode {
+        override val position = null
+        override val parentGuid = null
+        override val title = null
+    }
```
segunfamisa: I think this challenges the abstraction. If all fields are null here, it makes me wonder:

1. Is it that separators may have non-null values?
2. I don't imagine a separator should have a title. So, that means that title should not be in the base interface

mobile/android/android-components/components/concept/storage-bookmarks/src/main/java/mozilla/components/concept/storage/bookmarks/BookmarksStorage.kt
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Inserts bookmarks into storage.
+ */
+fun interface BookmarkInserter {
+    /**
+     * Inserts the given bookmark node into storage.
+     *
+     * @param node The [InsertableBookmarkNode] to insert.
+     */
+    suspend fun insert(node: InsertableBookmarkNode)
```
segunfamisa: I think we should be using `Result` APIs with our suspend functions. Will this throw an exception if insertion fails?

segunfamisa (approved):
I approve this, because I don't want to block the work, but I think the comments are worth taking a second look at.

segunfamisa:
Can we run a try?

boek:
[try](https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod-2025&landoCommitID=36772)

boek:
[try](https://treeherder.mozilla.org/jobs?repo=try&revision=81949d328d10c7236bd7758df2b0812ece3394cf)

PR #239, opened 2026-04-30
Bug 2035652 - Create basic scaffolding for conversational modal

MatthewTighe:
There's a bit of a problem with the keyboard pushing the text up, but I ran out of time and that feels pretty nonessential until we get further down the road.

image

### DIFF
--- a/mobile/android/android-components/components/feature/ask-page/src/main/java/mozilla/components/feature/ask/page/AskPageMiddleware.kt
+++ b/mobile/android/android-components/components/feature/ask-page/src/main/java/mozilla/components/feature/ask/page/AskPageMiddleware.kt
@@ -1,0 +1,75 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.components.feature.ask.page
+
+import kotlinx.coroutines.CoroutineDispatcher
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.flow.onCompletion
+import kotlinx.coroutines.launch
+import mozilla.components.concept.llm.CloudLlmProvider
+import mozilla.components.concept.llm.Llm
+import mozilla.components.feature.ask.page.ext.mapToRichDocument
+import mozilla.components.lib.state.Middleware
+import mozilla.components.lib.state.Store
+
+class AskPageMiddleware(
+    private val llmProvider: CloudLlmProvider,
+    private val scope: CoroutineScope,
+    private val dispatcher: CoroutineDispatcher = Dispatchers.Default,
+) : Middleware<AskPageState, AskPageAction> {
+
+    override fun invoke(
+        store: Store<AskPageState, AskPageAction>,
+        next: (AskPageAction) -> Unit,
+        action: AskPageAction,
+    ) {
+        val prevState = store.state
+        next(action)
+        when (action) {
+            is ViewAppeared -> scope.launch { observeProvider(store) }
+            is LlmProviderAction.ProviderAvailable -> scope.launch { llmProvider.prepare() }
+            is LlmProviderAction.ProviderReady -> {
+                val wasWaiting = prevState is AskPageState.WaitingToSendMessage
+                if (wasWaiting) {
+                    val ready = store.state as? AskPageState.Ready ?: return
+                    scope.launch { sendPrompt(store, ready.llm, ready.messages) }
+                }
+            }
+            is UserMessageSubmitted -> {
+                val ready = store.state as? AskPageState.Ready ?: return
+                scope.launch { sendPrompt(store, ready.llm, ready.messages) }
+            }
+            else -> {}
+        }
+    }
+
+    private suspend fun sendPrompt(
+        store: Store<AskPageState, AskPageAction>,
+        llm: Llm,
+        messages: List<Llm.Message>,
+    ) {
+        runCatching {
+            llm.prompt(Llm.ContextWindow(messages))
+                .mapToRichDocument(dispatcher)
+                .onCompletion { if (it == null) store.dispatch(ResponseCompleted) }
+                .collect { (markdown, document) ->
+                    store.dispatch(ReceivedParsedResponse(PartialResponse(markdown, document)))
+                }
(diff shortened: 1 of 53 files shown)


MatthewTighe:
try for the first patch: https://treeherder.mozilla.org/jobs?repo=try&revision=1e4614d2f004730022cfb909733a62fb8c5d4fa3

for the second: https://treeherder.mozilla.org/jobs?repo=try&revision=e36c39af213fc19d7a76f6f3525e8ed0d814e878

mobile/android/android-components/components/feature/ask-page/src/main/java/mozilla/components/feature/ask/page/AskPageMiddleware.kt
```diff
@@ -0,0 +1,44 @@
...
+) : Middleware<AskPageState, AskPageAction> {
+
+    override fun invoke(
+        store: Store<AskPageState, AskPageAction>,
+        next: (AskPageAction) -> Unit,
+        action: AskPageAction,
+    ) {
+        next(action)
+        if (action is UserMessageSubmitted) {
+            scope.launch { sendMessage(store, action.text) }
```
segunfamisa, line 30: should we trim the text?

mobile/android/android-components/components/feature/ask-page/src/main/java/mozilla/components/feature/ask/page/AskPageState.kt
```diff
@@ -0,0 +1,60 @@
...
+import mozilla.components.lib.state.State
+import mozilla.components.ui.richtext.ir.RichDocument
+
+/**
+ * An in-progress assistant response.
+ *
+ * @param rawMarkdown The accumulated raw markdown text received so far.
+ * @param richDocument The incrementally-parsed [RichDocument] for display.
+ */
+data class PartialResponse(val rawMarkdown: String, val richDocument: RichDocument)
```
segunfamisa, line 17: I wonder if it's going to be helpful to have a way to deserialize `RichDocument` to text, whenever we need to, instead of keeping two copies of roughly the same content in memory.

mobile/android/android-components/components/feature/ask-page/src/main/java/mozilla/components/feature/ask/page/AskPageReducer.kt
```diff
@@ -0,0 +1,45 @@
...
+        is AskPageState.Waiting -> AskPageState.Ready(messages = state.messages)
+        is AskPageState.Receiving -> AskPageState.Ready(
+            messages = state.messages + Llm.Message.Assistant(state.pendingResponse.rawMarkdown),
+        )
+        else -> state
+    }
+    is ResponseFailed -> when (state) {
+        is AskPageState.Waiting -> AskPageState.Ready(messages = state.messages, hasError = true)
+        is AskPageState.Receiving -> AskPageState.Ready(messages = state.messages, hasError = true)
+        else -> state
```
segunfamisa, line 43: is there a situation where the previous LLM call fails, and we don't want to set the ready state and flag that it has an error?

mobile/android/android-components/components/lib/llm-harness/src/main/java/mozilla/components/lib/llm/harness/LlmSessionImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+            }
+        }
+    }
+
+    override suspend fun switchProvider(provider: LlmProvider): Boolean {
+        require(provider in config.providers) { "Provider not in this session's provider list" }
+
+        if (isUnavailable(provider)) return false
+
+        if (config.preparationStrategy == LlmPreparationStrategy.ActiveOnly) {
```
segunfamisa, line 108: I think this block could benefit from some comments. Do I understand it correctly that, if we have a strategy to only prepare the active one, when the user switches, we have to prepare this?

---
Since the preparation strategy does not sound like something that would change in the course of a session, if a user switches providers back and forth, it seems like we would always attempt to prepare the new provider.

---
Would it be helpful, instead, if we checked whether the provider is prepared or not, and only then, do we prepare it?

mobile/android/android-components/components/lib/llm-harness/src/main/java/mozilla/components/lib/llm/harness/LlmSessionImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+    @OptIn(ExperimentalCoroutinesApi::class)
+    private suspend fun getLlm(): Llm =
+        awaitLlm(_activeProvider.value) ?: when (config.failoverStrategy) {
+            LlmFailoverStrategy.FailFast ->
+                throw Llm.Exception.unknown("No LLM provider is available")
+            LlmFailoverStrategy.AutoFailover, LlmFailoverStrategy.ManualFailover ->
+                _activeProvider
+                    .flatMapLatest { provider -> flow { emit(awaitLlm(provider)) } }
+                    .filterNotNull()
+                    .first()
```
segunfamisa, line 129: I'm not sure I really understand what this block is doing.

Am I reading this correctly: If we are not able to get an LLM from the `activeProvider`, we try again with the same active provider?

Is it a retry mechanism?

Reading the idea of a failover, I imagined it would iterate over the list of providers in the configuration and then pick the first successful one?

mobile/android/android-components/components/lib/llm-harness/src/main/java/mozilla/components/lib/llm/harness/LlmSessionImpl.kt
```diff
@@ -0,0 +1,136 @@
...
+import mozilla.components.concept.llm.CloudLlmProvider
+import mozilla.components.concept.llm.Llm
+import mozilla.components.concept.llm.LlmFailoverStrategy
+import mozilla.components.concept.llm.LlmPreparationStrategy
+import mozilla.components.concept.llm.LlmProvider
+import mozilla.components.concept.llm.LlmSession
+import mozilla.components.concept.llm.LlmSessionConfig
+import mozilla.components.concept.llm.LocalLlmProvider
+
+internal class LlmSessionImpl(private val config: LlmSessionConfig) : LlmSession {
```
segunfamisa, line 30: I think we could add some tests for this class. Idk what the plans for this implementation are beyond the prototype. Do we plan to revisit to add tests and all?

segunfamisa (approved):
Overall, the changes make sense to me. I had a couple of questions to help me understand the work stream.

I approved it, but I'd take that lightly, since I am just building context on the work.

PR #240, opened 2026-04-30
Bug 2030079: Fix intermittent test failure.

pollym:
This test was non-deterministic - it went:
- place food
- have the fox eat the food
- respawn the food in a random place
- check that the food is in a different place now

But there was a chance the food would respawn in the same place again (: This can be fixed by initially placing the food outside of the area where it can spawn. There is a "safe zone" for that right up against the walls, because food doesn't naturally spawn there (it makes the game quite annoying to play if it does)


pollym:
[here is a try](https://treeherder.mozilla.org/jobs?repo=try&revision=6f968b64533f4d21987574ad2b7eaa788d6a1906)

AndreyZIno (approved):
Looks good!

PR #242, opened 2026-04-30
Bug 2030040: deal with web notification issue.

pollym:
According to Claude:

> Without this call, the Bundle uses the boot ClassLoader during unparceling — which doesn't know about app or GeckoView classes.


pollym:
[here is a try](https://treeherder.mozilla.org/jobs?repo=try&revision=52c51651203548c938da3c075fa5904c602fcc54)

pollym:
doesn't look like this is actually required from the bugzilla discussion - it doesn't fix anything or usefully change any behaviour. so i'll close this.

PR #243, opened 2026-05-05
Bug 2000191: fix intermittent test failure caused by race condition.

pollym:
Root cause: [Bug 1879332](https://bugzilla.mozilla.org/show_bug.cgi?id=1879332) changed `TabSessionState.isActive()` to use `maxOf(lastAccess, createdAt)`.
Since createTab() defaulted `createdAt = System.currentTimeMillis()`, newly created test tabs have createdAt ≈ now. 
With `maxActiveTime = 0`, the check `now - createdAt <= 0` only returns false when at least 1ms has elapsed, making the result dependent on test execution speed.

Fix: Pass `createdAt = lastAccess` to all `createTab()` calls in tests that use `maxActiveTime = 0` and expect `inactive = true`. 
Since lastAccess values like 123L are epoch-old timestamps, `now - maxOf(123L, 123L)` is always a large positive number, so `isActive(0)` deterministically returns `false` regardless of when the test runs.


pollym:
[try is here](https://treeherder.mozilla.org/jobs?repo=try&revision=ff21e3bf6267175fbfc74cd8fb6b691ffd780fea)

titooan (approved):
Thank you for fixing that!

PR #249, opened 2026-05-08
Bug 2036474 - Updated s2s telemetry

fmasalha:


mobile/android/android-components/components/lib/llm-mlpa/src/main/java/mozilla/components/lib/llm/mlpa/MlpaLlmProvider.kt
```diff
@@ -37,7 +38,11 @@ class MlpaLlmProvider(
     val storage: MlpaTokenStorage,
     val mlpaService: MlpaService,
 ) : CloudLlmProvider {
-    override val info = LlmProvider.Info(nameRes = R.string.mlpa_llm_provider_name, iconRes = R.drawable.firefox_icon)
+    override val info = LlmProvider.Info(
+        nameRes = R.string.mlpa_llm_provider_name,
+        iconRes = R.drawable.firefox_icon,
+        model = ModelID.mozSummarization.value,
```
segunfamisa: I think that if we have a future where we have multiple model id's, or a dynamic model id, this would break.

If I remember correctly, we have the model id as part of the request.......so I don't know if it is "static". I believe it could be dynamic, so far...based on the design we have.

segunfamisa: I believe that we have a fixed value we define here - https://github.com/mozilla-firefox/firefox/blob/f63224e3cb2a50ad8d2082f71c23f73407ee620f/mobile/android/android-components/components/lib/llm-mlpa/src/main/java/mozilla/components/lib/llm/mlpa/MlpaLlm.kt#L26-L34

Is there a way we could pass the model from the llmProvider to the Llm Request? So that we are certain that the correct model owned by the provider is the same one used for the prompt request

boek:
Since this is a telemetry patch let's move it to phabricator

PR #255, opened 2026-05-14
Bug 2037736 - Added a store to the settings and implemented storeProvider for persisted state

fmasalha:
https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod-2025&landoCommitID=46701

### DIFF
--- a/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
+++ b/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
@@ -41,6 +41,8 @@ import mozilla.components.concept.sync.AuthType
 import mozilla.components.concept.sync.OAuthAccount
 import mozilla.components.concept.sync.Profile
 import mozilla.components.feature.addons.ui.AddonFilePicker
+import mozilla.components.feature.summarize.settings.SummarizationSettings
+import mozilla.components.lib.state.helpers.StoreProvider.Companion.navBackStackStore
 import mozilla.components.service.fxrelay.eligibility.Eligible
 import mozilla.components.support.base.feature.ViewBoundFeatureWrapper
 import mozilla.components.support.ktx.android.view.showKeyboard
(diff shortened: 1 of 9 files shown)


mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/SettingsFragment.kt
```diff
@@ -83,10 +93,107 @@ import org.mozilla.fenix.snackbar.FenixSnackbarDelegate
...
+    )
+    is SettingsAction.PageSummariesSettingsLoaded -> state.copy(
+        summarizeSettingsState = state.summarizeSettingsState.copy(
+            isFeatureEnabled = action.isFeatureEnabled,
+            isGestureEnabled = action.isGestureEnabled,
+        ),
+    )
+    SettingsAction.SettingsViewCreated -> state
+}
+
```
segunfamisa: I think we should put these in different files - Middleware, State, etc. We should not put them here

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/pagesummaries/PageSummariesSettingsFragment.kt
MatthewTighe, line 71: I don't think this is your problem to solve, but just FYI I'm not super happy that we have to have two stores here. This should work, since the "parent" store maintains the state, but it would be good for us as a team to invest in a richer solution at some point.

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AIControlsScreen.kt
MatthewTighe, line 87: Can we file a follow-up to migrate this screen to a Store? I think we've had enough examples of working around the complexity here that it's probably worth just aligning on the common pattern

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AIControlsFragment.kt
MatthewTighe, line 64: Curious to know: do we need to remember this controller or the callbacks object for a specific reason?

MatthewTighe (approved):
Nice, had a question and a couple thoughts but nothing blocking. LGTM! thanks for handling this one

MatthewTighe:
It does look like the test failures in the try are real though, could you investigate that to make sure?

PR #256, opened 2026-05-14
Bug 2039662 - Added a preload on summary settings on summarization

fmasalha:


fmasalha:
https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod-2025&landoCommitID=46703

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationFragment.kt
```diff
@@ -183,17 +185,20 @@ class SummarizationFragment : BottomSheetDialogFragment() {
...
+                        settings = summarizeSettings,
+                        onLearnMoreClicked = { openLearnMoreLink() },
+                        storeViewModel.viewModelScope,
+                    ),
                 ),
-            ),
-        )
+            )
+        }
+        LaunchedEffect(Unit) { settingsStore.dispatch(SettingsPreLoaded) }
```
MatthewTighe, line 201: I don't think this `Action` accurately describes an event that has occurred in the system - it seems more like we are wanting to _start_ the process of pre-loading the settings. 

Overall, this approach seems like it's reinforcing the code smell of having two stores here. I think the ideal situation we want is to allow the `SummarizationStore` to be able to react to `ViewAppeared` and have that result in an updated `SettingsState` that is delivered to the settings screen. 

This could be a good discussion for team time tomorrow I think. I don't feel strongly that we _have_ to address it as part of this patch, but I think this would be a good motivation for doing so.

As an aside, could you update the bug to better describe the problem we are trying to solve here rather than the steps of the solution? The problem is relatively clear from the code, but it would be good to have a digestible reference for future readers.

MatthewTighe (changes requested):

PR #257, opened 2026-05-14
Bug 2039671 - Use navigation action IDs instead of destination IDs in AIFeatureMetadataDestination

fmasalha:
https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod-2025&landoCommitID=46707

### DIFF
--- a/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AIFeatureMetadataDestination.kt
+++ b/mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AIFeatureMetadataDestination.kt
@@ -36,11 +36,11 @@ data class Destination(val id: Int, override val label: Int) : AIFeatureMetadata
  */
 val AIFeatureMetadata.destination: AIFeatureMetadataDestination? get() = when (id) {
     PageSummaryFeature.id -> Destination(
-        id = R.id.pageSummariesSettingsFragment,
+        id = R.id.action_aiControlsFragment_to_pageSummariesSettingsFragment,
         label = R.string.ai_controls_more_page_summary_settings,
     )
     TranslationsAIControllableFeature.id -> Destination(
-        id = R.id.translations_settings_graph,
+        id = R.id.action_aiControlsFragment_to_translationsSettingsFragment,
         label = R.string.ai_controls_more_translations_settings,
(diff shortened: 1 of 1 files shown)


mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AIFeatureMetadataDestination.kt
```diff
@@ -36,11 +36,11 @@ data class Destination(val id: Int, override val label: Int) : AIFeatureMetadata
  */
 val AIFeatureMetadata.destination: AIFeatureMetadataDestination? get() = when (id) {
     PageSummaryFeature.id -> Destination(
-        id = R.id.pageSummariesSettingsFragment,
+        id = R.id.action_aiControlsFragment_to_pageSummariesSettingsFragment,
```
segunfamisa, line 39: Question. Did this work before? What is the difference between using the destination & using the navigation action id?

segunfamisa: while I don't think this would help us here, but to prevent the possibility of adding an unexpected type of "int",

could we add something like this in `Destination`

```kotlin
data class Destination(
  @param:IdRes val id: Int, 
  @param:StringRes override val label: Int
) : AIFeatureMetadataDestination
```

It wouldn't help us here, because both `R.id.pageSummariesSettingsFragment` and `R.id.action_...` are both `id` resources, but the string one can help us avoid future issues there.

segunfamisa (approved):

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/settings/ai/AIFeatureMetadataDestination.kt
fmasalha: > Question. Did this work before? What is the difference between using the destination & using the navigation action id?

From my understanding, putting in the defined action uses the xml action defined with the animation that most of the rest of settings uses. Before it was just doing a navigation to the fragment rather than using the navgraph. We originally had it using the actions but I believe it switched when we moved the nav link to the metadata class.

PR #264, opened 2026-05-21
Bug 2039562: Fix initial state flicker.

pollym:
Also added some ui tests around the new game screen. The flicker is handled by alpha-ing out the components if they're null & it's not currently possible through the compose test apis to make assertions about the alpha state, so that part was verified visually, but we can check that the other behaviours of ui components are wired up as expected. Did a small refactor around this to hoist state and expose smaller more testable params to NewGameScreen.

### DIFF
--- a/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
+++ b/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
@@ -66,6 +66,10 @@ fun LongFoxGameScreen() {
         }
         val restartGame = { gameState = GameState(numCells = numCells, size = Size(canvasSizePx, canvasSizePx)) }
         SideEffect {
+            // this is to satisfy the IDE.
+            // the warning seems to be a false positive with compose state
+            // https://youtrack.jetbrains.com/projects/KT/issues/KT-78881/K2-False-positive-Assigned-value-is-never-read-in-composable-function
+            @Suppress("AssignedValueIsNeverRead")
             if (gameState.shouldCelebrateScore && !celebrationShown) celebrationSeed = gameState.score
             celebrationShown = gameState.shouldCelebrateScore
         }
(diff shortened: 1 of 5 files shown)


mobile/android/fenix/app/longfox/src/test/resources/robolectric.properties
```diff
@@ -0,0 +1,7 @@
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+sdk=35
```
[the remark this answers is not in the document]
pollym: gosh

pollym:
[try, with a new added line break](https://treeherder.mozilla.org/jobs?repo=try&revision=c2ac5a66b4f5ca201ea96f37e5534bb81daca874)

mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
```diff
@@ -66,6 +66,7 @@ fun LongFoxGameScreen() {
         }
         val restartGame = { gameState = GameState(numCells = numCells, size = Size(canvasSizePx, canvasSizePx)) }
         SideEffect {
+            @Suppress("AssignedValueIsNeverRead")
```
segunfamisa, line 72: We could put a comment here explaining that this is to satisfy the IDE. But effectively, the warning seems to be a fasle positive with compose state - https://youtrack.jetbrains.com/projects/KT/issues/KT-78881/K2-False-positive-Assigned-value-is-never-read-in-composable-function

segunfamisa (approved):
Thanks for this patch!

We reviewed together and talked about:
1. Future iterations in the `NewGameScreen` could possibly contain a "NewGameScreenState" 
2. Ideas for the testing - if we want to keep exploring how to test the alpha visibility, we could inspect the modifiers in a node using the `node.layoutInfo.getModifierInfo()` API and then inspect the modifiers applied

Overall, this looks good and other improvements can be sent as follow-ups

mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxGameScreen.kt
pollym: i will do this! thank you :)

PR #273, opened 2026-05-29
Bug 2040784 - Introduce AIFeatureState to handle the "Unknown" states of AI controls

segunfamisa:

### DIFF
--- a/mobile/android/android-components/components/lib/ai-controls/src/main/java/mozilla/components/lib/ai/controls/AIFeatureRegistry.kt
+++ b/mobile/android/android-components/components/lib/ai-controls/src/main/java/mozilla/components/lib/ai/controls/AIFeatureRegistry.kt
@@ -4,14 +4,32 @@
 
 package mozilla.components.lib.ai.controls
 
+import android.content.Context
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.launch
 import mozilla.components.concept.ai.controls.AIControllableFeature
 import mozilla.components.concept.ai.controls.AIFeatureMetadata
 import mozilla.components.concept.ai.controls.AIFeatureRegistry
+import mozilla.components.concept.ai.controls.AIFeatureState
 
 /**
- * Creates the default implementation of [AIFeatureRegistry], which enforces unique feature IDs.
+ * Creates the implementation of [AIFeatureRegistry], which enforces unique feature IDs.
  */
-fun AIFeatureRegistry.Companion.default() = object : AIFeatureRegistry {
+fun AIFeatureRegistry.Companion.default(
+    scope: CoroutineScope,
+    context: Context,
+): AIFeatureRegistry =
+    DefaultAIFeatureRegistry(scope, AIFeatureBlockStorage.dataStore(context))
+
+/**
+ * Default implementation of [AIFeatureRegistry] that enforces unique feature IDs and
+ * initializes feature states based on the block status stored in [AIFeatureBlockStorage].
+ */
+internal class DefaultAIFeatureRegistry(
+    private val scope: CoroutineScope,
+    private val storage: AIFeatureBlockStorage,
+) : AIFeatureRegistry {
     // LinkedHashMap allows us to maintain the order for later use.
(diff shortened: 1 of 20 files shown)


segunfamisa:
try: https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod-2025&landoCommitID=51795

mobile/android/android-components/components/lib/ai-controls/src/main/java/mozilla/components/lib/ai/controls/AIFeatureRegistry.kt
```diff
@@ -4,21 +4,47 @@
...
- * Creates the default implementation of [AIFeatureRegistry], which enforces unique feature IDs.
+ * Creates the implementation of [AIFeatureRegistry], which enforces unique feature IDs.
  */
-fun AIFeatureRegistry.Companion.default() = object : AIFeatureRegistry {
+fun AIFeatureRegistry.Companion.default(
+    scope: CoroutineScope,
+    context: Context,
+): AIFeatureRegistry =
+    DefaultAIFeatureRegistry(scope, AIFeatureBlockStorage.dataStore(context))
+
```
MatthewTighe, line 24: Can we swap the context dependency for an `AIFeatureBlockStorage` dependency?

mobile/android/android-components/components/lib/ai-controls/src/test/java/mozilla/components/lib/ai/controls/AIFeatureBlockTest.kt
```diff
@@ -100,4 +101,14 @@ class AIFeatureBlockTest {
         block.unblock()
         assertFalse(block.isBlocked.first())
     }
+
+    private class FakeAIFeatureRegistry : AIFeatureRegistry {
```
MatthewTighe, line 105: I think we could just reuse  `AIFeatureRegistry.inMemory` for this

mobile/android/android-components/components/lib/ai-controls/src/test/java/mozilla/components/lib/ai/controls/AIFeatureRegistryTest.kt
```diff
@@ -21,4 +44,77 @@ class AIFeatureRegistryTest {
...
+    private class TestAIFeatureBlockStorage(initialBlocked: Boolean) : AIFeatureBlockStorage {
+        val blockedFlow = MutableStateFlow(initialBlocked)
+        override val isBlocked: Flow<Boolean> = blockedFlow
+
+        override suspend fun setBlocked(isBlocked: Boolean) {
+            blockedFlow.value = isBlocked
+        }
+    }
+
+    private class TestAIControllableFeature(
```
MatthewTighe, line 106: These could also use the `.inMemory` factories I believe

mobile/android/android-components/components/lib/ai-controls/src/main/java/mozilla/components/lib/ai/controls/AIFeatureRegistry.kt
```diff
@@ -4,21 +4,47 @@
...
             "AI feature with id=${feature.id} is already registered"
         }
+
+        scope.launch {
+            if (feature.featureState.first() is AIFeatureState.Unknown) {
+                val aiFeaturesBlocked = storage.isBlocked.first()
+                feature.set(!aiFeaturesBlocked)
+            }
+        }
+
```
MatthewTighe, line 47: We should move to the IO dispatcher here since we are using the `MainScope`

MatthewTighe (approved):
few small things, but this otherwise LGTM. thanks for taking this one

mobile/android/android-components/components/lib/ai-controls/src/main/java/mozilla/components/lib/ai/controls/AIFeatureRegistry.kt
MatthewTighe: Ah, I see why this was chosen now - we didn't want to expose the internal type. Carry on

PR #274, opened 2026-05-29
Bug 2040784 - Introduce AIFeatureState to handle the "Unknown" states…

MatthewTighe:
… of AI controls

I commandeered this from Segun. Only downstream changes from me were to swap to the IO dispatcher while collecting the storage flow in the registry and a cosmetic change to remove the text fixtures he introduced in favor of the existing `inMemory` factories.


boek (approved):
This looks good to me. Is it easy to check this fix in the s2s case to verify the CFR isn't shown?

PR #277, opened 2026-06-01
Bug 2040784 - Introduce AIFeatureState to handle the "Unknown" states…

MatthewTighe:
… of AI controls


boek (approved):

MatthewTighe:
The last land failed because I forgot to push a test fix I had locally, but here is another try just in case

https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod-2025&landoCommitID=52554

PR #279, opened 2026-06-04
Bug 2039564: Add nimbus flag to enable longfox.

pollym:
Defaulted to false for nightly but true for developers because i think we all need a bit more joy in our lives quite honestly.

[Here is a try](https://treeherder.mozilla.org/jobs?repo=try&revision=64629a31ba5a52bbaee269410003d70774fa26b3)


AndreyZIno (approved):
Looks good!

PR #289, opened 2026-06-12
Bug 2043181: Switch between longfox and privacy dashboard cta

pollym:
Bug 2043181: Switch between longfox and privacy dashboard cta

- if longfox is enabled, the privacy button cta launches longfox, otherwise it launches the privacy dashboard
-- previously longfox was only accessible from the peeking fox animation. i have renamed the logic around the peeking fox animation to clarify that this is no longer the same thing as the longfox entry point.
- send a telemetry event every time the longfox cta is shown (i had left this out of https://bugzilla.mozilla.org/show_bug.cgi?id=2039559 until we had figured out what the cta was!)


pollym:
[here is a try for all 3 commits](https://treeherder.mozilla.org/jobs?repo=try&revision=e3dd860c98af0ff2bbd61f6432717ad54e39745e)

mcarare (approved):
LGTM!

PR #290, opened 2026-06-12
Bug 2045749: Add game in new dialog to prevent talkback issues.

pollym:
I've also changed the feature API to pass in `Context` instead of the container `ViewGroup`.
I don't love passing context as a parameter, but it is the thing we need to spawn a new dialog, so it makes more sense than passing in a `ViewGroup` and getting the `Context` from it.

[there is a try here](https://treeherder.mozilla.org/jobs?repo=try&revision=eddb7ec074a0856f90d1a1c2fa4ca10c4a6db99d)

### DIFF
--- a/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxFeature.kt
+++ b/mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxFeature.kt
@@ -6,11 +6,16 @@
 
 package org.mozilla.fenix.longfox
 
+import android.content.Context
+import android.content.ContextWrapper
 import android.view.ViewGroup
+import androidx.activity.ComponentDialog
 import androidx.activity.compose.BackHandler
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.ui.platform.ComposeView
 import androidx.compose.ui.platform.ViewCompositionStrategy
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
 import org.mozilla.fenix.longfox.GleanMetrics.Longfox
 
 /**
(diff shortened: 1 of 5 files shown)


mobile/android/fenix/app/longfox/src/main/kotlin/org/mozilla/fenix/longfox/LongFoxFeature.kt
```diff
@@ -37,27 +42,49 @@ interface LongFoxFeatureApi {
...
                     BackHandler {
-                        container.removeView(this)
-                        disposeComposition()
+                        dialog.dismiss()
                     }
                 }
             },
         )
+
+        // On destroy, dismiss the window with the host so it can't leak if the activity goes away mid-game.
```
segunfamisa, line 76: 👌🏾

segunfamisa (approved):
Code-wise, this looks good to me. I don't know why this fixed it though, or what other options could have helped here - I wonder if requesting focus for the game would bring the screen reader's focus there and it won't get stuck in the background?

Either way, this looks good to me, and thanks for handling the activity clean up!

pollym:
> Code-wise, this looks good to me. I don't know why this fixed it though, or what other options could have helped here - I wonder if requesting focus for the game would bring the screen reader's focus there and it won't get stuck in the background?
> 
> Either way, this looks good to me, and thanks for handling the activity clean up!

that's fair. i would like to spend a bit of time looking at alternatives here for sure, there are probably nicer ways to handle it!
i probably didn't explain very well... the accessibility issue was happening because there were a bunch of other views that were siblings of the game composeview in the view hierarchy. it basically gets added into whatever view you pass in. this was a bit of a hangover from when i was playing around with putting it in smaller-than-fullscreen views, so you could embed it wherever you liked. 
but we don't actually do this now, it's a full screen overlay. So making it a dialog makes sense, and it fixes the issue because it's shown on top of the other content rather than as a sibling. even though you couldn't see the other content visually, they were still visible for traversal for accessibility services. 
claude suggested making the other views invisible for accessibility and then visible again when you dismiss the game, but i didn't really like that approach, made me nervous that we would miss the re-visible-ising 😅

PR #291, opened 2026-06-12
Bug 2040618: Fix touch slop shenanigans.

pollym:
It turns out touch slop isn't the right thing to use in this situation to differentiate between a tap and a swipe. It's the minimum distance to begin a drag, not a cutoff between a tap and swipe. So i've put the dp magic number back in! sorry andrey (:


pollym:
[here is a try](https://treeherder.mozilla.org/jobs?repo=try&revision=6927274dd480d678a0f3599b24b56f6d3ce8daa0)

mcarare (approved):

PR #293, opened 2026-06-16
Bug 2047940: Refine longfox CTA.

pollym:
When the button is going to launch longfox, changed the button text to "help catch trackers" and the animated TypewriterText to "tap to play!" These are the best strings we have available for 153. Hopefully they make it a bit clearer what is going to happen when you hit the button. Additionally, trimmed the padding underneath: this was a hangover from when the animated text was a CTA to launch longfox. We don't need that any more and it's pushing the other home page content down.

Before

image
image

After

after-no-peek
after-with-peek


pollym:
[here is a try](https://treeherder.mozilla.org/jobs?repo=try&revision=64a7326412cce0befa44da80e5e315f341484886)

moyheen (approved):
LGTM!

segunfamisa (approved):

PR #297, opened 2026-06-17
Bug 2048200: Expose longfox secret setting to all release types.

pollym:
This means that if people are in the longfox experiment and hate it, they can turn it off. 
conversely, if you are longfox curious you could switch it on!


moyheen (approved):
The code looks good, thanks for making the changes!

pollym:
behaviour tweaked slightly following chat with ux.

- removed the longfox switch from the homepage settings, so there is only one switch.
-- this switch shows / hides the button
-- this switch has a label which matches the button type: either "privacy report" or, if longfox is enabled, "help catch trackers"

- there is still an "enable longfox" switch in secret settings, 
-- this switch will change the behaviour of the button
-- so qa can test the feature and if people are desperate to play the game but aren't in the experiment there is a way to do that. Or, if they are in the experiment but want the privacy dashboard, they can also flip this setting.

this is a much smaller simpler change :)
[try is running here](https://treeherder.mozilla.org/jobs?repo=try&revision=38e19fd4dc63a8996c48c98368d4b7591a109b1d)

PR #303, opened 2026-06-23
Bug 2047013 - Allows rating of summarizations

boek:
image

### DIFF
--- a/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationReducer.kt
+++ b/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationReducer.kt
@@ -24,11 +24,18 @@ fun summarizationReducer(state: SummarizationState, action: SummarizationAction)
     is SummarizationFailed -> SummarizationState.Error(SummarizationError.SummarizationFailed(action.exception))
     is ReceivedParsedDocument -> state.updateDocument(action.document)
     is SettingsClicked -> when (state) {
-        is SummarizationState.Summarized -> SummarizationState.Settings(info = state.info, document = state.document)
+        is SummarizationState.Summarized ->
+            SummarizationState.Settings(info = state.info, document = state.document, feedback = state.feedback)
         else -> state
     }
     is SettingsBackClicked -> when (state) {
-        is SummarizationState.Settings -> SummarizationState.Summarized(info = state.info, document = state.document)
+        is SummarizationState.Settings ->
+            SummarizationState.Summarized(info = state.info, document = state.document, feedback = state.feedback)
+        else -> state
+    }
+    is SummaryFeedbackProvided -> when (state) {
+        is SummarizationState.Summarized ->
+            if (state.feedback == null) state.copy(feedback = action.feedback) else state
         else -> state
     }
     else -> state
--- a/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationState.kt
+++ b/mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationState.kt
@@ -57,10 +57,13 @@ sealed class SummarizationState : State {
      *
      * @param info metadata about the LLM that generated the summary
      * @param document The generated document.
+     * @param feedback The rating the user gave the summary, or `null` if they have not rated it.
+     *  The rating is single-use: once set it is never cleared, disabling further feedback.
      */
     data class Summarized(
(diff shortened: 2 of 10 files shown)


boek:
When you rate: 
image

xianghongai:
https://addons.mozilla.org/en-US/firefox/ `Log in`:

```json
{
  "detail": "You do not have permission to perform this action."
}
```

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/ui/SummaryContentLoaded.kt
nicholaspoon03, line 49:
```suggestion
 * @param onFeedbackClicked Invoked when the user rates the summary. When `null`, the feedback
```

nicholaspoon03, line 75: nit: can we add some spacing between each of these composables to make it easier to read

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationReducer.kt
nicholaspoon03, line 38: Hmm can you explain why we have this if branch? I'm probably missing something but I don't really get why we need it

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationTelemetryMiddleware.kt
nicholaspoon03, line 86: I think it could make sense for this to be here. It could also make sense to add an entirely new middleware for this, since we seem to just be leveraging our telemetry system to measure this non telemetry event. Let me know what your thoughts are. If you think it should be here, that works for me too

nicholaspoon03, line 161: If the user presses one of the two buttons and then changes their mind, it seems that we will record their response multiple times. Should we do something to account for this?

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/SummarizationState.kt
nicholaspoon03, line 122: nit: It seems like we use the term good review / bad review and positive / negative. Can we centralize on 1 naming so its less confusing

mobile/android/fenix/app/src/main/java/org/mozilla/fenix/summarization/SummarizationTelemetryMiddleware.kt
nicholaspoon03, line 163: I'm wondering if we should just do `feedback.name.lowercase()` here

mobile/android/android-components/components/feature/summarize/src/main/java/mozilla/components/feature/summarize/ui/SummaryContentLoaded.kt
nicholaspoon03, line 202: I tried using talkback with the buttons. Thinking aloud, would it be more clear to make the content description have some sort of verb so its more like a call to action. I think something like `Rate good summary` would be good? Right now, users swipe from `AI can make mistakes` to `Good summary, button` - maybe this also depends on the placement, FWIW, which you were also skeptical about.

After commenting this, I also tested the buttons google search gemini and they say `Good response, button`, `Bad response, button` - so maybe it's more about the positioning then and not the content description

nicholaspoon03, line 231: It's a bit weird because if the user gives a rating, the content description becomes `Bad summary, Button, disabled` and `Good summary, Button, disabled`. I would also expect for the fact that a response was already selected to be announced somehow

nicholaspoon03 (commented):
Hey Jeff,
Good work! Just a few things that I wanted to ask about and I don't know how data collection processes work on github - so I asked in slack to see how to do that

flodolo:
Stumbled upon this patch by accident. 

Patches with strings should not go through this workflow for now?

DatGamet (approved):
lgtm

CherrieTheShifter:
Yes This

boek:
closing this in favor of https://phabricator.services.mozilla.com/D317863

PR #315, opened 2026-07-02
Bug 1996675 - Add homepage/shortcuts_library source to top sites telemetry. r=gl

devotaaabel:


mobile/android/fenix/app/metrics.yaml
```diff
@@ -9645,7 +9676,9 @@ top_sites:
       A user clicked a Contile top site
     extra_keys:
       source:
-        description: The source of the event, example "newtab", "urlbar"
+        description: |
+          The surface where the Contile top site was clicked.
+          One of "homepage" or "shortcuts_library".
```
gabrielluong, line 9765: For the contile telemetry, since we already use "newtab" to refer to "homepage", let's not change that because it would affect the existing mapping and reporting for the sponsored top sites if we don't give them a heads up about this change.

You could add "shortcuts_library", but I don't want us to mess up the existing telemetry reporting. For this, could you add a new enum class? 

Separately, I will ask them if we will still need this telemetry going forward since telemetry is also reported through the Ads Client, but proceed with the intention that we still need this.

gabrielluong (approved):

mobile/android/fenix/app/metrics.yaml
devotaaabel: We chatted about this offline, and we're going to keep it as-is and just make the ads team aware of the change.

devotaaabel:
https://treeherder.mozilla.org/jobs?repo=try&landoInstance=lando-prod-2025&landoCommitID=65443

PR #321, opened 2026-07-30
Bug 2059390: fix crash in SitePermissionsFragment.onDismiss

pollym:
If a fragment is detached, it is no longer safe to call `requireComponents`, because this method throws if the `context` is null. In this case, `BaseBrowserFragment` is not attached and returning true from `shouldHide` will stop us from showing the permission prompt, rather than crashing. Also checked `hidePermissionsPrompt` which will be called if `shouldHide` is true. This allows for the case of the fragment not being attached by using an optional and a let, so should be safe to call in this detached fragment case.

[try is running here](https://treeherder.mozilla.org/jobs?repo=try&revision=8dbfc878c87991d2da9167e8ad5daed2bb54579a)


pollym:
i couldn't think of an easy way to write a unit test around this, but if anyone has an idea lmk

segunfamisa (approved):
Thanks for this patch!

PR #323, opened 2026-08-06
Bug 2061436: Protect StorageStatsManager from IOException. r?segun

moztcampbell:
We are seeing odd crashes querying storage for some telemetry. Swallow the exception to avoid crashes in this telemetry path.


moztcampbell:
This is an extension of your work in https://bugzilla.mozilla.org/show_bug.cgi?id=2008785 to also handle `IOException`

PR #324, opened 2026-08-07
Bug 2061436: Protect StorageStatsManager from IOException. r=segun

moztcampbell:
We are seeing odd crashes querying storage for some telemetry. Swallow the exception to avoid crashes in this telemetry path.


segunfamisa (approved):
Oh interesting. Thanks for this. I didn't know we were getting I/O exception here too. I was trying to be conservative about the swallowed exceptions, but this looks nice, thanks.

PR #325, opened 2026-08-07
Bug 2044409 - Filter slider updates while Accessibility shouldUseAutoSize is false

moztcampbell:
The engine code doesn't allow setting a fontSizeFactor when auto-sizing is enabled. We previously tried to disable the Slider that updates the size factor in order to prevent update messages, but the disable doesn't take effect until next recomposition. As a result we see crashes in monkey testing which adds noise even if it is highly unlikely for a normal user to hit.


ohall-m (approved):
Thanks!

moztcampbell:
[testing-other] This is a crash that was only triggered by monkey testing and would be clumsy to test in a repeatable way.

shtrom:
Why, thanks @lando-web... Better late than never, huh?

PR #329, opened 2026-08-14
Bug 2063644 - Remove obsolete nsFrame references from comments. r?#layout

aethanyc:
We removed `nsFrame` in bug 1630704. The code now lives in `nsIFrame`.

### DIFF
--- a/layout/generic/nsIFrame.h
+++ b/layout/generic/nsIFrame.h
@@ -39,11 +39,6 @@
 #  define MAX_REFLOW_DEPTH 1026
 #endif
 
-/* nsIFrame is in the process of being deCOMtaminated, i.e., this file is
-   eventually going to be eliminated, and all callers will use nsFrame instead.
-   At the moment we're midway through this process, so you will see inlined
-   functions and member variables in this file.  -dwh */
-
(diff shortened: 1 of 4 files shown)


layout/generic/nsIFrame.h
```diff
@@ -39,11 +39,6 @@
 #  define MAX_REFLOW_DEPTH 1026
 #endif
 
-/* nsIFrame is in the process of being deCOMtaminated, i.e., this file is
```
dshin-moz, line 42: I'm very curious about the history behind this comment. Was there meant to be a replacement, or did we just rename the replacement nsIFrame?

dshin-moz (approved):
Seems simple enough, kinda curious about the history of `nsFrame` (Not a blocker of course)

aethanyc:
Part of the history https://bugzilla.mozilla.org/show_bug.cgi?id=190735

aethanyc:
I think `nsIFrame` was meant to be an abstract interface, and `nsFrame` (or other concrete classes) can implement it. But it has been always `nsFrame` be the only implementation, and we declared `nsIFrame` be the private API in gecko only.

PR #333, opened 2026-08-16
Bug 2015253 - Only return a container frame top level docs during print preview. r=tnikkel,#layout

emilio:
This matches the pre-regression logic, and fixes the coordinate system
of the slicing fragmentation fallback.

Unfortunately testing this is kind of a PITA, because the obvious way to
do it would be WPT print reftests, but that requires e10s. Regular
reftests don't have any print stuff that would reliably test the
fragmentation fallback I think?

I can try to come up with something tho...


emilio:
I'm testing the github workflow, ignore this one, let's do phabricator since @tnikkel is not in the pilot thing.

CherrieTheShifter:
Yes This

PR #335, opened 2026-08-18
Bug 2058496 - Deal with SVG context paint properly in border-image fast path. r=#layout

emilio:


layout/painting/nsImageRenderer.cpp
```diff
@@ -1020,9 +1020,11 @@ ImgDrawResult nsImageRenderer::DrawBorderImageComponent(
         nsLayoutUtils::GetSamplingFilterForFrame(mForFrame);
 
     if (!RequiresScaling(aFill, aHFill, aVFill, aUnitSize)) {
+      SVGImageContext svgContext;
+      SVGImageContext::MaybeStoreContextPaint(svgContext, mForFrame, subImage);
```
dshin-moz, line 1024: Kinda unfortunate naming - Maybe storing contextpaint isn't the only thing it does...

dshin-moz (approved):

layout/painting/nsImageRenderer.cpp
emilio: Well, if you think about it the embedder's color-scheme is part of the context... But yeah, agreed.

PR #337, opened 2026-08-18
Bug 2064471 - Make -moz-appearance a legacy shorthand of appearance. r#style

emilio:
Don't restrict values for now.


servo/components/style/properties/shorthands.toml
```diff
@@ -242,13 +242,11 @@ spec = "https://drafts.fxtf.org/motion-1/#offset-shorthand"
 
 [-webkit-perspective]
 sub_properties = ["perspective"]
-derive_serialize = true
```
dshin-moz, line 245: So it seems that we can't have "IS_LEGACY_SHORTHAND" + `derive_serialize = true` anymore? Worth commenting that?

dshin-moz (approved):
Seems we're skipping what feels like a superfluous derive and expecting `parse_legacy` and such. Nice.

servo/components/style/properties/shorthands.toml
emilio: I guess we could and instead deal with it on the expansion instead of expecting `to_css_legacy`. I actually like that a bit more.

emilio:
Rebased + fixed the property exposure / `property_database.js`

PR #339, opened 2026-08-19
Bug 2063854: Don't try to add anchor frame by the same name multiple times. r=#layout-reviewers

dshin-moz:


layout/base/PresShell.cpp
```diff
@@ -11697,8 +11696,22 @@ void PresShell::AddAnchorPosAnchorImpl(const nsAtom* aName, nsIFrame* aFrame,
   entry.InsertElementAt(matchOrInsertionIdx, aFrame);
 }
 
-void PresShell::AddAnchorPosAnchor(const nsAtom* aName, nsIFrame* aFrame) {
-  AddAnchorPosAnchorImpl(aName, aFrame, /* aForMerge = */ false);
+void PresShell::AddAnchorPosAnchor(const Span<const StyleAtom>& aNames, nsIFrame* aFrame) {
+  nsTHashSet<const nsAtom*> added{static_cast<uint32_t>(aNames.Length())};
```
dshin-moz: (Probably can get away with array here, I imagine)

layout/base/PresShell.h
```diff
@@ -816,8 +817,11 @@ class PresShell final : public nsStubDocumentObserver,
                                const nsIFrame* aPositionedFrame) const;
   void CollectAnchorNames(const nsIFrame* aPositionedFrame,
                           nsTArray<nsString>& aResult);
-  void AddAnchorPosAnchor(const nsAtom* aName, nsIFrame* aFrame);
+  void AddAnchorPosAnchor(const Span<const StyleAtom>& aNames, nsIFrame* aFrame);
```
emilio: Pass `Span<>` by value.

layout/base/PresShell.cpp
emilio: Yeah I suspect you could just an AutoTArray or so.

emilio (approved):

layout/base/PresShell.h
```diff
@@ -816,8 +817,11 @@ class PresShell final : public nsStubDocumentObserver,
                                const nsIFrame* aPositionedFrame) const;
   void CollectAnchorNames(const nsIFrame* aPositionedFrame,
                           nsTArray<nsString>& aResult);
-  void AddAnchorPosAnchor(const nsAtom* aName, nsIFrame* aFrame);
+  void AddAnchorPosAnchor(Span<const StyleAtom> aNames, nsIFrame* aFrame);
+  void RemoveAnchorPosAnchor(Span<const StyleAtom> aNames, nsIFrame* aFrame);
+private:
   void RemoveAnchorPosAnchor(const nsAtom* aName, nsIFrame* aFrame);
```
aethanyc: Nit: maybe rename this to `RemoveAnchorPosAnchorImpl` and move the declaration after `AddAnchorPosAnchorImpl` in the private section. 

https://searchfox.org/firefox-main/rev/ad5db99e4d2e00e9b5db2700c69050d709c7f340/layout/base/PresShell.h#1955-1956

dshin-moz: Hmmm, tbh, I'd rather lean on the overload in this case - Given that 1) # params is the same, 2) implicit-casting an atom into a span by accident feels improbable.

... I'll move it to the private grouping, though

aethanyc (approved):

layout/base/PresShell.h
aethanyc: That works, too. Thanks!

PR #343, opened 2026-08-20
Bug 2041870: Don't let mUseAccessibilityTheme override mUseDocumentColors. r=firefox-style-system-reviewers

dshin-moz:
See comment in bug - mUseDocumentColors already takes mUseAccessibilityTheme
into account if the pref is set to do so.

---

Lando: [link](https://lando.moz.tools/pulls/firefox-autoland/343/)
Bugzilla: [bug 2041870](https://bugzilla.mozilla.org/show_bug.cgi?id=2041870)

:no_entry_sign: This pull request has 3 blockers.

### DIFF
--- a/layout/style/nsMediaFeatures.cpp
+++ b/layout/style/nsMediaFeatures.cpp
@@ -326,7 +326,7 @@ StylePrefersContrast Gecko_MediaFeatures_PrefersContrast(
     return StylePrefersContrast::NoPreference;
   }
   const auto& prefs = PreferenceSheet::PrefsFor(*aDocument);
-  if (!prefs.mUseAccessibilityTheme && prefs.mUseDocumentColors) {
+  if (prefs.mUseDocumentColors) {
     return StylePrefersContrast::NoPreference;
   }
   const auto& colors = prefs.ColorsFor(ColorScheme::Light);


emilio (changes requested):
I think this seems wrong. This basically disables `prefers-contrast` altogether, right?

`prefers-contrast` is intended to match even when not using high-contrast mode. If we need an override for that, it needs to be added on top.

emilio (changes requested):
(See above)

emilio:
(testing, but I still think this is wrong)

emilio (changes requested):
(See above)

OS-AC713 (changes requested):
This is a regression. You've removed the accessibility theme check entirely, which means that users with visual impairments (e.g., color blindness, low contrast requirements) will no longer have their dedicated theme respected. Now document colors will always override accessibility settings, breaking the primary purpose of the mUseAccessibilityTheme flag — which exists specifically to protect readability for these users.

This is not a fix; it's a functional regression that directly harms the accessibility features Firefox is supposed to provide. Please revert this change and resubmit with a proper logic that preserves the priority of mUseAccessibilityTheme over document colors, or explain why we should accept this trade-off — but I don't see any valid reason.

CherrieTheShifter:
Yes This

emilio:
@dshin-moz should we close this?

PR #347, opened 2026-08-21
Bug 2065660 - Remove redundant atomization in attribute selector parsing. r=#style

emilio:
If the name is lowercase there's no need to atomize twice.

### DIFF
--- a/servo/components/selectors/parser.rs
+++ b/servo/components/selectors/parser.rs
@@ -3165,7 +3165,7 @@ where
 
     let attribute_flags = parse_attribute_flags(input)?;
     let value = value.as_ref().into();
-    let local_name_lower;
+    let local_name_lower: Impl::LocalName;
     let local_name_is_ascii_lowercase;
     let case_sensitivity;
     {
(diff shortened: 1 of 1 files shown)


servo/components/selectors/parser.rs
```diff
@@ -3165,7 +3165,7 @@ where
 
     let attribute_flags = parse_attribute_flags(input)?;
     let value = value.as_ref().into();
-    let local_name_lower;
+    let local_name_lower: Impl::LocalName;
     let local_name_is_ascii_lowercase;
     let case_sensitivity;
     {
```
dshin-moz, line 3171: Nit: tuple assignment (Weak preference, maybe `Impl::LocalName` to avoid type deduction issues)?

dshin-moz (approved):

servo/components/selectors/parser.rs
emilio: Yeah, but it gets a bit weird because some of the temporaries escape and some don't...

PR #352, opened 2026-08-31
Bug 2058181 - More carefully clamp grid tracks. r=#layout

emilio:
---

Lando: [link](https://lando.moz.tools/pulls/firefox-autoland/352/)
Bugzilla: [bug 2058181](https://bugzilla.mozilla.org/show_bug.cgi?id=2058181)

:no_entry_sign: This pull request has 2 blockers.

### DIFF
--- a/layout/generic/nsGridContainerFrame.cpp
+++ b/layout/generic/nsGridContainerFrame.cpp
@@ -62,6 +62,8 @@ static mozilla::LazyLogModule gGridContainerLog("GridContainer");
 // http://drafts.csswg.org/css-grid/#overlarge-grids
 static const int32_t kMaxLine = 10000;
 static const int32_t kMinLine = -10000;
+static const int32_t kMaxTrack = kMaxLine - 1;
+
 // The maximum line number, in the zero-based translated grid.
 static const uint32_t kTranslatedMaxLine = uint32_t(kMaxLine - kMinLine);
 static const uint32_t kAutoLine = kTranslatedMaxLine + 3457U;
@@ -1510,19 +1512,18 @@ struct nsGridContainerFrame::TrackSizingFunctions {
     if (!aIsSubgrid) {
       ExpandNonRepeatAutoTracks();
     }
-
-#ifdef DEBUG
     if (mHasRepeatAuto) {
+      if (mRepeatAutoStart > kMaxTrack) {
+        // The `repeat(auto-fill/fit)` track is outside the clamped grid.
+        mHasRepeatAuto = false;
+      }
       MOZ_ASSERT(mExpandedTracks.Length() >= 1);
-      const unsigned maxTrack = kMaxLine - 1;
-      // If the exanded tracks are out of range of the maximum track, we
+      // If the expanded tracks are out of range of the maximum track, we
       // can't compare the repeat-auto start. It will be removed later during
       // grid item placement in that situation.
-      if (mExpandedTracks.Length() < maxTrack) {
-        MOZ_ASSERT(mRepeatAutoStart < mExpandedTracks.Length());
-      }
+      MOZ_ASSERT_IF(mExpandedTracks.Length() < kMaxTrack,
+                    mRepeatAutoStart < mExpandedTracks.Length());
     }
-#endif
   }
 
  public:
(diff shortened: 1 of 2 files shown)


layout/generic/nsGridContainerFrame.cpp
```diff
@@ -1510,19 +1512,21 @@ struct nsGridContainerFrame::TrackSizingFunctions {
...
-#ifdef DEBUG
     if (mHasRepeatAuto) {
+      if (mRepeatAutoStart > kMaxTrack) {
+        // The `repeat(auto-fill/fit)` track is outside the clamped grid.
+        mHasRepeatAuto = false;
+      }
+#ifdef DEBUG
       MOZ_ASSERT(mExpandedTracks.Length() >= 1);
-      const unsigned maxTrack = kMaxLine - 1;
       // If the exanded tracks are out of range of the maximum track, we
```
dholbert: maybe fix the typo while you're here (s/exanded/expanded/)

dshin-moz: (Previous typo - ex*p*anded)

layout/generic/nsGridContainerFrame.cpp
```diff
@@ -1510,19 +1512,21 @@ struct nsGridContainerFrame::TrackSizingFunctions {
...
       // can't compare the repeat-auto start. It will be removed later during
       // grid item placement in that situation.
-      if (mExpandedTracks.Length() < maxTrack) {
+      if (mExpandedTracks.Length() < kMaxTrack) {
```
dshin-moz: Maybe worth `MOZ_ASSERT_IF`

layout/generic/nsGridContainerFrame.cpp
```diff
@@ -1510,19 +1512,21 @@ struct nsGridContainerFrame::TrackSizingFunctions {
...
     if (mHasRepeatAuto) {
+      if (mRepeatAutoStart > kMaxTrack) {
+        // The `repeat(auto-fill/fit)` track is outside the clamped grid.
+        mHasRepeatAuto = false;
+      }
+#ifdef DEBUG
       MOZ_ASSERT(mExpandedTracks.Length() >= 1);
-      const unsigned maxTrack = kMaxLine - 1;
       // If the exanded tracks are out of range of the maximum track, we
       // can't compare the repeat-auto start. It will be removed later during
```
dshin-moz, line 1522: FWIW, it should probably say "dropped" instead of "removed" now

layout/generic/nsGridContainerFrame.cpp
```diff
@@ -2052,16 +2075,20 @@ class MOZ_STACK_CLASS nsGridContainerFrame::LineNameMap {
   void ExpandRepeatLineNamesForSubgrid(
       const StyleGenericLineNameList<StyleInteger>& aStyleLineNameList) {
     const auto& lineNameList = aStyleLineNameList.line_names.AsSpan();
-    const uint32_t maxCount = mClampMaxLine + 1;
-    const uint32_t end = lineNameList.Length();
-    for (uint32_t i = 0; i < end && mExpandedLineNames.Length() < maxCount;
-         ++i) {
-      const auto& item = lineNameList[i];
+    SmallPointerArray<const NameList> names;
+    auto AppendExpandedNames = [&]() -> bool {
```
dshin-moz, line 2077: Couldn't this be factored out like `EmplaceExpandedTrack`?

dholbert (approved):
Thanks!

layout/generic/nsGridContainerFrame.cpp
dholbert: (ah right, dshin caught this as well; midair'ed on the review, didn't notice until after submitting)

layout/generic/nsGridContainerFrame.cpp
```diff
@@ -1510,19 +1512,21 @@ struct nsGridContainerFrame::TrackSizingFunctions {
     if (!aIsSubgrid) {
       ExpandNonRepeatAutoTracks();
     }
-
-#ifdef DEBUG
     if (mHasRepeatAuto) {
+      if (mRepeatAutoStart > kMaxTrack) {
+        // The `repeat(auto-fill/fit)` track is outside the clamped grid.
+        mHasRepeatAuto = false;
```
dshin-moz, line 1518: So wouldn't this give up on sizing for whatever is inside? I guess we're already truncating content, but `ExpandNonRepeatAutoTracks` crash in the bug should no longer happen with exiting early without extending the grid then truncating it?

dshin-moz (approved):

layout/generic/nsGridContainerFrame.cpp
emilio: Not easily because we really want to reuse the local...

layout/generic/nsGridContainerFrame.cpp
emilio: Not changing behavior tho, I'm just moving the change from the end of Expand... to here (its only caller), so that I can early-return.

dshin-moz (approved):

emilio:
Fixed an embarrasing bug (`mRepeatAutoStart = mExpandedTracks.Length();` used to obviously run before appending)

emilio:
See https://bugzilla.mozilla.org/show_bug.cgi?id=2067996

PR #357, opened 2026-09-04
Bug 2066814 - Restrict the properties that highlight pseudos get at cascade time. r=#style

emilio:
Matches Chrome and the spec in terms of em / lh units etc, and prevents
border-radius to set the border-background flag without extra tracking.

---

Lando: [link](https://lando.moz.tools/pulls/firefox-autoland/357/)
Bugzilla: [bug 2066814](https://bugzilla.mozilla.org/show_bug.cgi?id=2066814)

:no_entry_sign: This pull request has 1 blocker.

### DIFF
--- a/servo/components/style/properties/data.py
+++ b/servo/components/style/properties/data.py
@@ -1202,7 +1202,7 @@ class PropertyRestrictions:
 
     @staticmethod
     def webkit_text_properties():
-        props = set(
+        return set(
             [
                 # Kinda like css-text?
                 "-webkit-text-stroke-width",
@@ -1210,7 +1210,6 @@ class PropertyRestrictions:
                 "-webkit-text-stroke-color",
             ]
         )
-        return props
 
     # https://drafts.csswg.org/css-pseudo/#first-letter-styling
     @staticmethod
@@ -1365,6 +1364,17 @@ class PropertyRestrictions:
             + PropertyRestrictions.shorthand(data, "outline")
             + PropertyRestrictions.shorthand(data, "font")
             + PropertyRestrictions.shorthand(data, "font-synthesis")
+        )
+
+    # https://drafts.csswg.org/css-pseudo/#highlight-pseudos
+    # https://drafts.csswg.org/css-pseudo/#highlight-styling
+    def highlight(data):
+        return set(
+            [
+                "color",
+                "background-color",
+            ]
+            + PropertyRestrictions.spec(data, "css-text-decor")
         )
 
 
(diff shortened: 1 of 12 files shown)


servo/components/style/properties/data.py
```diff
@@ -1367,6 +1366,19 @@ def cue(data):
...
+    # https://drafts.csswg.org/css-pseudo/#highlight-styling
+    def highlight(data):
+        props = set(
+            [
+                "color",
+                "background-color",
+            ]
+            + PropertyRestrictions.spec(data, "css-text-decor")
+        )
+        props = props.union(PropertyRestrictions.webkit_text_properties())
```
dshin-moz: Should these be allowed?

> Vendor-prefixed properties such as [-webkit-text-fill-color](https://compat.spec.whatwg.org/#propdef--webkit-text-fill-color) are not applicable to the [highlight pseudo-elements](https://drafts.csswg.org/css-pseudo/#highlight-pseudo-element).

and https://wpt.live/css/css-pseudo/highlight-styling-003.html

servo/components/style/properties/data.py
```diff
@@ -1367,6 +1366,19 @@ def cue(data):
...
 
+    # https://drafts.csswg.org/css-pseudo/#highlight-pseudos
+    # https://drafts.csswg.org/css-pseudo/#highlight-styling
+    def highlight(data):
+        props = set(
+            [
+                "color",
+                "background-color",
+            ]
+            + PropertyRestrictions.spec(data, "css-text-decor")
```
dshin-moz, line 1377: Imagine we don't yet support highlight applying to svg texts?

dshin-moz (changes requested):
Don't think webkit text properties are allowed, are they? 
Otherwise the change looks good - seems to be mostly refactoring `Option<PropertyFlags>` to be just `PropertyFlags`, since that's a bitfield and can be empty

servo/components/style/properties/data.py
emilio: That test doesn't stop passing, but I guess that's just because we don't look at those properties during rendering.

servo/components/style/properties/data.py
emilio: I assume this question is because of https://wpt.live/css/css-pseudo/textpath-selection-011.html? I haven't debugged but it seems we don't look at `::selection` in `SVGTextFrame`, so checks out.

servo/components/style/properties/data.py
emilio: (For context, I removed this)

dshin-moz (approved):

PR #360, opened 2026-09-09
Bug 2070311 - Fix interaction of rule cache reset reuse and visited optimization. r=#style

emilio:
We're hitting the "used cached reset properties" path, and thus we don't
notice that .item indeed sets border-color, which is a visited dependent
property. So we end up incorrectly with `border-color: currentcolor` in
the computed style, and thus wrong rendering.

Track the "used visited dependent properties" with a style bit and carry
that around instead.

---

Lando: [link](https://lando.moz.tools/pulls/firefox-autoland/360/)
Bugzilla: [bug 2070311](https://bugzilla.mozilla.org/show_bug.cgi?id=2070311)

:warning: This pull request has 3 warnings.
:no_entry_sign: This pull request has 1 blocker.

### DIFF
--- a/servo/components/style/properties/cascade.rs
+++ b/servo/components/style/properties/cascade.rs
@@ -245,9 +245,6 @@ pub enum CascadeMode<'a, 'b> {
     Visited {
         /// The cascade for our unvisited style.
         unvisited_context: &'a computed::Context<'b>,
-        /// The properties set on the unvisited style. These are useful if we can prove that the
-        /// visited rules are the same as the unvisited declarations.
-        unvisited_properties: &'a LonghandIdSet,
     },
 }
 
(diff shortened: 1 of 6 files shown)


servo/components/style/properties/cascade.rs
dshin-moz, line 393: using cached reset props here

dshin-moz, line 396: border properties aren't inherited, so excluded from here

servo/components/style/properties/cascade.rs
```diff
@@ -363,16 +357,18 @@ where
             // It also wouldn't be super-profitable, only a handful :visited properties are
             // non-inherited.
             using_cached_reset_properties = false;
-            let visited_dependent_props = LonghandIdSet::visited_dependent();
             // If we match the same rules when visited and unvisited, and we know there isn't any
             // visited-dependent properties, we can avoid gathering the declarations, we know none
             // would be relevant.
             if unvisited_context.builder.rules.as_ref() != Some(rules)
-                || unvisited_properties.contains_any(visited_dependent_props)
```
dshin-moz, line 371: `border-color` never gets seen here, so we skip over this

dshin-moz (approved):

